PR #5, opened 2020-06-24
Fix field name typo

thegreatape:


Mubelotix:
It is strange to see that my previous comment has been moved here. That's a bug IMO.

PR #22, opened 2020-07-24
Replace Vec references with slices (&Vec<T> to &[T])

### LINKED ISSUE #21: Use slices (&[T]) instead of references to a vector (&Vec<T>) in the documents API
Currently the documents API on master takes `&Vec<T>` parameters when adding or deleting documents.  Slices (`&[T]`) are more idiomatic while providing the same functionality.  Making the parameter a slice allows more flexibility for the users of the API, and can allow the API to be used with arrays and other data structures besides `Vec`s.

I can also make a PR with this change if that would be helpful.

Aeledfyr:
Fixes #21

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ async fn main() {
     let mut books = client.get_or_create("books").await.unwrap();
 
     // Add some books in the index
-    books.add_documents(&vec![
+    books.add_documents(&[
         Book{book_id: 123,  title: String::from("Pride and Prejudice")},
         Book{book_id: 456,  title: String::from("Le Petit Prince")},
         Book{book_id: 1,    title: String::from("Alice In Wonderland")},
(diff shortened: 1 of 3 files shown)


Aeledfyr:
`cargo test -- --test-threads=1` passes all 50 tests for me.  In theory this may not be an API breaking change from the `&Vec<T>` parameters, since `&Vec<T>` can deref into `&[T]`. 

I can't find any other avoidable instances of `Vec<T>` being used as a parameter.  It would be nice to reduce the amount of allocation in the settings API, but then you'd have to deal with tons of lifetimes and have two separate structs for the owned an unowned variants (from deserializing).

Aeledfyr:
Whoops, sorry about that. I rebased off of master, but didn't realize that would dismiss the review.

PR #32, opened 2020-07-30
Generated README

### LINKED ISSUE #27: Generate the README
Following this comment https://github.com/meilisearch/meilisearch-rust/issues/9#issuecomment-665285593

Would be better the #9 was closed before doing this one so that we know the README template we want.

curquiza:
Closes #27

### DIFF
--- a/.github/scripts/check-release.sh
+++ b/.github/scripts/check-release.sh
@@ -2,10 +2,15 @@
 
 # Checking if current tag matches the package version
 current_tag=$(echo $GITHUB_REF | tr -d 'refs/tags/v')
-file_tag=$(grep '^version = ' Cargo.toml | cut -d '=' -f 2 | tr -d '"' | tr -d ' ')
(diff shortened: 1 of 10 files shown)


curquiza:
Hum! Didn't realized that! Thanks @Mubelotix!
I moved the documentation part from `src/lib.rs` to `README.tpl` because it's more convenient for us (the maintainers) to change a `.tpl` than the comments in the `src/lib.rs` file... Is there a way to use the README of the repo in docs.rs?

Maybe @MarinPostma, @LegendreM or @Kerollmops, you would have an idea to solve that?

curquiza:
@Mubelotix, I moved back the installation part in the `src/lib.rs`. That's too bad because we lose the feature of `cargo readme` that generates the current version with `{{version}}` but we do not have any choice 🙂
I let the other parts in the main README since it's **really** for us (the maintainers) to change a `.tpl` than comments in `.rs` file!
Are you ok with that now? 🙂

PR #36, opened 2020-07-30
Improve error handling and move from MeiliSearch 0.11 to MeiliSearch 0.12 errors.

### LINKED ISSUE #23: Support Meilisearch v0.12 errors
Currently, this library uses string matching on the JSON errors to determine the error type, as Meilisearch only returned a message, and no error code.  Meilisearch v0.12 adds an `errorCode` field to the error JSON, which should allow for better forwards compatibility with the future, in case of changes in the human readable message.  Since the current method relies entirely on string parsing, adding new fields to the JSON error breaks it.  I have a fixed implementation here, but there are many error types that need to be added.
```rust
impl From<&serde_json::Value> for Error {
    fn from(message: &serde_json::Value) -> Error {
        // Error codes from https://github.com/meilisearch/MeiliSearch/blob/v0.12.0/meilisearch-error/src/lib.rs
        match message.get("errorCode").and_then(|v| v.as_str()) {
            Some("index_not_found") => Error::IndexNotFound,
            Some("index_already_exists") => Error::IndexAlreadyExist,
            Some("invalid_index_uid") => Error::InvalidIndexUid,
            Some("missing_primary_key") => Error::CantInferPrimaryKey,
            Some("maintenance") => Error::ServerInMaintenance,
            Some(_) => Error::Unknown(message.to_string()),
            None => {
                // Meilisearch 0.11 and below 
                match message.get("message").and_then(|v| v.as_str()) {
                    Some("Impossible to create index; index already exists") => Error::IndexAlreadyExist,
                    Some("Could not infer a primary key") => Error::CantInferPrimaryKey,
                    Some(m) if m.starts_with("Server is in maintenance, please try again later") => Error::ServerInMaintenance,
                    Some(m) if m.starts_with("Index ") && m.ends_with(" not found") => Error::IndexNotFound,
                    Some(m) if m.starts_with("Index must have a valid uid;") => Error::InvalidIndexUid,
                    _ => {
                        Error::Unknown(message.to_string())
                    },
                }
            }
        }
    }
}
```
I am using https://github.com/meilisearch/MeiliSearch/blob/v0.12.0/meilisearch-error/src/lib.rs as a source for the error codes, as the Meilisearch documentation does not currently have a list.

Aeledfyr:
This is a rewrite of the errors module, so it has breaking changes, and this only supports MeiliSearch 0.12 errors, so this removes support for MeiliSearch 0.11. (Fixes #23)

The error codes and error types have documentation comments loosely based around the official error code documentation (https://docs.meilisearch.com/errors/), which is incorrect in some places.  The documentation is missing the `index_not_found` error code, and has a `missing_header` error code that only existed between MeiliSearch 0.10 and 0.11 and was never released.

This changes the `HttpError` to contain a String on `wasm` so that the `Error::Unknown` variant could be removed, but a more descriptive error enum would be a better than a string.

All of the error enums (`Error`, `ErrorCode`, and `ErrorType`) are documented and marked as `#[non_exhaustive]`, although the `ErrorType` enum doesn't seem like it would change.

I have run clippy and all of the tests as specified in the contributing guidelines.  (Although the docker command for running the latest server doesn't work, as the `--no-analytics` argument requires a parameter after it.)

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -134,10 +134,13 @@ fn parse_response<Output: DeserializeOwned>(
             }
             Err(e) => {
                 error!("Request succeed but failed to parse response");
-                return Err(Error::from(e.to_string().as_str()));
+                return Err(Error::Unknown(e.to_string()));
             }
         };
     }
     warn!("Expected response code {}, got {}", expected_status_code, status_code);
-    Err(Error::from(body.as_str()))
+    match from_str(&body) {
+        Ok(e) => Err(Error::from(&e)),
+        Err(e) => Err(Error::Unknown(e.to_string()))
+    }
 }
--- a/src/request.rs
+++ b/src/request.rs
@@ -100,12 +100,12 @@ pub(crate) async fn request<Input: Serialize + std::fmt::Debug, Output: 'static 
             Ok(text) => text,
             Err(e) => {
                 error!("Invalid response: {:?}", e);
-                return Err(Error::Unknown("Invalid response".to_string()));
+                return Err(Error::HttpError("Invalid response".to_string()));
             }
         }
         Err(e) => {
             error!("Invalid response: {:?}", e);
-            return Err(Error::Unknown("Invalid response".to_string()));
+            return Err(Error::HttpError("Invalid response".to_string()));
         }
     };
 
@@ -117,7 +117,7 @@ pub(crate) async fn request<Input: Serialize + std::fmt::Debug, Output: 'static 
         }
     } else {
         error!("Invalid response");
-        Err(Error::Unknown("Invalid utf8".to_string()))
+        Err(Error::HttpError("Invalid utf8".to_string()))
     }
 }
 
--- a/src/request.rs
+++ b/src/request.rs
@@ -134,13 +134,13 @@ fn parse_response<Output: DeserializeOwned>(
             }
             Err(e) => {
                 error!("Request succeed but failed to parse response");
-                return Err(Error::Unknown(e.to_string()));
+                return Err(Error::ParseError(e));
             }
         };
     }
     warn!("Expected response code {}, got {}", expected_status_code, status_code);
     match from_str(&body) {
         Ok(e) => Err(Error::from(&e)),
-        Err(e) => Err(Error::Unknown(e.to_string()))
+        Err(e) => Err(Error::ParseError(e))
     }
 }
--- a/src/request.rs
+++ b/src/request.rs
@@ -91,8 +91,8 @@ pub(crate) async fn request<Input: Serialize + std::fmt::Debug, Output: 'static 
         Ok(response) => Response::from(response),
         Err(e) => {
             error!("Network error: {:?}", e);
-            return Err(Error::UnreachableServer)
-        },
+            return Err(Error::UnreachableServer);
+        }
     };
     let status = response.status() as u16;
     let text = match response.text() {
@@ -102,7 +102,7 @@ pub(crate) async fn request<Input: Serialize + std::fmt::Debug, Output: 'static 
                 error!("Invalid response: {:?}", e);
                 return Err(Error::HttpError("Invalid response".to_string()));
             }
-        }
+        },
         Err(e) => {
             error!("Invalid response: {:?}", e);
             return Err(Error::HttpError("Invalid response".to_string()));
@@ -141,6 +141,6 @@ fn parse_response<Output: DeserializeOwned>(
     warn!("Expected response code {}, got {}", expected_status_code, status_code);
     match from_str(&body) {
         Ok(e) => Err(Error::from(&e)),
-        Err(e) => Err(Error::ParseError(e))
+        Err(e) => Err(Error::ParseError(e)),
     }
 }
--- a/src/client.rs
+++ b/src/client.rs
@@ -182,16 +182,18 @@ impl<'a> Client<'a> {
     /// # Example
     ///
     /// ```
-    /// # use meilisearch_sdk::{client::*, indexes::*, errors::Error};
+    /// # use meilisearch_sdk::{client::*, indexes::*, errors::{Error, ErrorCode}};
     /// #
     /// # #[tokio::main]
     /// # async fn main() {
     /// let client = Client::new("http://localhost:7700", "masterKey");
     ///
     /// match client.get_health().await {
-    ///     Ok(()) => println!("server is operationnal"),
-    ///     Err(Error::ServerInMaintenance) => eprintln!("server is in maintenance"),
-    ///     _ => panic!("should never happen"),
+    ///     Ok(()) => println!("server is operational"),
+    ///     Err(Error::MeiliSearchError { error_code: ErrorCode::Maintenance, .. }) => {
+    ///         eprintln!("server is in maintenance")
+    ///     },
+    ///     Err(e) => panic!("should never happen: {}", e),
     /// }
     /// # }
     /// ```
@@ -203,7 +205,9 @@ impl<'a> Client<'a> {
             204,
         ).await;
         match r {
-            Err(Error::Unknown(m)) if &m == "null" => Ok(()),
+            // This shouldn't be an error; The status code is 200, but the request
+            // function only supports one successful error code for some reason
+            Err(Error::Empty) => Ok(()),
             e => e
         }
     }
@@ -236,7 +240,9 @@ impl<'a> Client<'a> {
             204,
         ).await;
         match r {
-            Err(Error::Unknown(m)) if &m == "null" => Ok(()),
+            // This shouldn't be an error; The status code is 200, but the request
+            // function only supports one successful error code for some reason
+            Err(Error::Empty) => Ok(()),
             e => e
         }
     }
--- a/src/client.rs
+++ b/src/client.rs
@@ -203,12 +203,13 @@ impl<'a> Client<'a> {
             self.apikey,
             Method::Get,
             204,
-        ).await;
+        )
+        .await;
         match r {
             // This shouldn't be an error; The status code is 200, but the request
             // function only supports one successful error code for some reason
             Err(Error::Empty) => Ok(()),
-            e => e
+            e => e,
         }
     }
 
@@ -238,12 +239,13 @@ impl<'a> Client<'a> {
             self.apikey,
             Method::Put(HealthBody { health }),
             204,
-        ).await;
+        )
+        .await;
         match r {
             // This shouldn't be an error; The status code is 200, but the request
             // function only supports one successful error code for some reason
             Err(Error::Empty) => Ok(()),
-            e => e
+            e => e,
         }
     }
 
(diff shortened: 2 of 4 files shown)


curquiza:
Thanks a loooooot for your work @Aeledfyr! And also thanks @Mubelotix for your review!

I'm requesting someone from the core-engine team as reviewer before merging 😁

Aeledfyr:
@Mubelotix The only remaining `Unknown` variant is `ErrorCode::Unkown`, which is limited to unhandled error codes from MeiliSearch.  I would recommend keeping it, as it allows users to get useful error codes and error messages if they use a more recent version of MeiliSearch, or if there are some error codes that we don't cover.  Without that variant, any unrecognized error codes from MeiliSearch would cause the error to be ignored or misinterpreted as a different error, preventing the user from getting the error message and link that was sent by MeiliSearch.

I can see the point to removing it, but I would recommend against removing it so that we don't have to add more error handling within the error handling, and so that the user can get a reasonable error in almost any case.

Mubelotix:
Yes I totally agree. I said Error::Unknown but I was thinking about Error::HttpError.

curquiza:
Hello @Aeledfyr!
Just to keep you informed: we do not forget you 😉
Most of our team is on vacation during summer or have a lot of work to do, so all is a little bit slower. It might take time to merge your PR, but be sure you will have a review from the core team and your PR will be merged!
Thanks again for your work and your patience!

Kerollmops:
Hey @Aeledfyr and @Mubelotix,

Thank you both for this SDK and this PR, it is quite good.

However, I am a little bit concerned about this `#[non_exhaustive]` and `ErrorCode::Unknown` mixing. The primary goal of the `#[non_exhaustive]` attriubute is to force the library user match on a last pattern (all other potential enum variant) even if all of the existing variants are matched by the `match`. It helps users bump the library version by ensuring that all variant are considered and no compilation error will raise (if all previous variants exist already).

```rust
match code {
  ... => (),
  _ => panic!("Wh0o0t?"),
}
```

The `ErrorCode::Unknown` variant is here to match all of the potential unknown error code returned by the engine, it kind of intersect with the `#[non_exhaustive]` attribute, the user will be forced to match on this variant anyway, so.

```rust
match code {
  ... => (),
  Unknown(code) => panic!("{}", code),
  _ => panic!("Wh0o0t?"),
}
```

But do we want this, do we want users to be able to bump the SDK crate version without taking care of the new errors variant?

Aeledfyr:
The original reason for the `ErrorCode::Unknown` variant is so that we can display a reasonable error in the case that the error code isn't handled.  It isn't intended to be matched on, as that could cause breaking changes:  It's main purpose is to give reasonable errors for code like this
```rust
match code {
    // snip...
    e => panic!("Error: {}", e),
}
```

I don't think that `#[non_exhaustive]` forces the user to handle all cases in your second example - the `_ => ...` pattern could also match the `ErrorCode::Unknown` variant, so it works in cases that the user doesn't explicitly match on the `Unknown` variant.  We could potentially prevent matching on the `Unknown` variant by wrapping the inner error string in a private wrapper and using `#[doc(hidden)]`, which allows us to retain the ability to display good error messages while preventing breaking changes from new error codes.
```rust
pub UnknownErrorCode(pub(crate) String);

pub enum ErrorCode {
    // ... snip
    #[doc(hidden)]
    Unknown(UnknownErrorCode),
}
```

As for the question of whether we should use `#[non_exhaustive]` or `Unknown` in the first place, adding new errors to the SDK should not be a breaking change.  That would require a major version change every time a new error code is added, which may not always line up with MeiliSearch's version increases.  That would also prevent this SDK from ever stabilizing with `semver` versioning, because even once `MeiliSearch` is stable, new features will likely have new error codes and supporting those would require breaking changes.

If we can guarantee that MeiliSearch treats adding error codes as a breaking change, we could do it, but I would still not recommend it.

Kerollmops:
Very good point @Aeledfyr, I would say that you are right in the way that we should keep both, the `#[non_exhautive]` attribute and the `Unknown` variant. I am also very interrested in moving the string error message into another type that wraps it, but I am not a partisan of the `#[doc(hidden]` types/variants.

I would say that introducing a `UnknownErrorCode` would prevent the user from matching on it and implementing `Display` and `Debug` would be enough for the user to print it if required (also `pub(crate)` doesn't seems required, as it is used and declared in the same file).

LegendreM:
Hello  @Aeledfyr and @Mubelotix,

I think that you could manage `MeilisearchError` implementing `serde` on it, you'll be able to drop all your `parse`/`as_str` functions, It would be easier to add new errors.

Moreover your `parse_response` function would be clearer with it:
```rust
fn parse_response<Output: DeserializeOwned>(
    body: String,
) -> Result<Output, Error> {

    #[derive(Deserialize)]
    #[serde(untagged)]
    enum ResponseWrapper<Output> {
        Valid(Output),
        Invalid(MeiliSearchError)
    }
    
    match from_str::<ResponseWrapper<Output>>(&body) {
        Ok(ResponseWrapper::Valid(output)) => {
            // trace!("Request succeed");
            Ok(output)
        }
        Ok(ResponseWrapper::Invalid(meilisearch_error)) => {
            // warn!("Expected response code {}, got {}", expected_status_code, status_code);
            Err(Error::MeiliSearchError(meilisearch_error))
        }
        Err(e) => {
            // error!("Request succeed but failed to parse response");
            Err(Error::ParseError(e))
        }
    }
}
```

see [here](https://gist.github.com/LegendreM/34ae001b41acc2632203b806fd3fc1a9) a gist of how I would do it.

Thanks for your work by the way 😃

Aeledfyr:
@Kerollmops Sounds good, implementing that now.

@LegendreM While I agree that using serde for the error types may be useful, it can't be used as easily as your example shows.

* The `ErrorCodeWrapper`, which is required to have serde's automatic deriving work with the `ErrorCode` while supporting the `Unknown` variant, makes matching on the error much harder.  Every place that would have been `ErrorCode::{Variant}` would now be `ErrorCodeWrapper::Known(ErrorCode::{Variant})`, which is too much friction for use in a library.  In this case the main benefit we get is a small decrease in development effort, but for an overall decrease in the usability of the library.
* The `as_str` function would have to stay to be able to properly implement `std::fmt::Display`.  Currently your implementation of `Display` for `ErrorCode` and `ErrorType` just calls `.to_string()`, which is just a wrapper around `Display`.  This will cause infinite recursion if either of the types is printed.
* The implementation of `parse_response` completely ignores the status codes of the response and instead bases whether the response succeeded entirely off of the json response.  In the case that `parse_response` is called with `Output` as `serde_json::Value`, it can only return an error if parsing failed:  An error response with a `400`/`500` status code would be treated as a success.  While the current implementation has issues of treating `204` as an error when `200` was expected, it works well enough — I believe that changing that is out of the scope of the PR.

I could move the `MeiliSearchError` variant out into it's own struct and implement `Serialize` and `Deserialize` on them, but that doesn't have as much flexibility and potential backwards/forwards compatibility as the current implementation of `From<&serde_json::Value>`.  It potentially could, but that would require a manual implementation of `Deserialize`, which ends up being an unmaintainable mess.

LegendreM:
@Aeledfyr my code was more an example of what could be done, I think that a good compromise could be found digging deeper in it. But the new implementation works well which make your PR mergable. 😄

LegendreM:
Hey @Aeledfyr, could you use `&str` instead of `String` at https://github.com/meilisearch/meilisearch-rust/pull/36/files#diff-2873e3b87f87572c686867a1c112e3fdR124-R128 ?

curquiza:
@Aeledfyr the tests are failing now because the new version of MeiliSearch is out (`[v0.13.0](https://github.com/meilisearch/MeiliSearch/releases/tag/v0.13.0)`) and the tests are running with the latest MeiliSearch version.

This PR #42 fixes the tests: all your tests should pass once we merge the #42 and you [rebase your branch](https://gist.github.com/curquiza/5f7ce615f85331f083cd467fc4e19398) 🙂

Aeledfyr:
@LegendreM That method should use `&str` instead of `String`, but I believe that that is out of the scope of this PR - this is for fixing error handling.  The requests and response parsing code has enough issues that it should be fixed or rewritten in a different PR, as changing it (especially changing the function signature) will require larger changes to the codebase.  So far I've tried to minimize the impact of the error rewrite on the rest of the library.

@curquiza Sounds good; should I squash the commits into one when I rebase?

curquiza:
> @curquiza Sounds good; should I squash the commits into one when I rebase?

As you wish, but in any case I will "[Squash and merge](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/merging-a-pull-request#merging-a-pull-request-on-github)" on the github interface when merging your PR 😉

Kerollmops:
@curquiza I would prefer you to merge by creating a _merge commit_, this way we would be able to keep each commit in its own and also be able to identify from which PR it comes from.

src/request.rs
```diff
@@ -91,21 +91,21 @@ pub(crate) async fn request<Input: Serialize + std::fmt::Debug, Output: 'static
...
+        }
     };
     let status = response.status() as u16;
     let text = match response.text() {
         Ok(text) => match JsFuture::from(text).await {
             Ok(text) => text,
             Err(e) => {
                 error!("Invalid response: {:?}", e);
-                return Err(Error::Unknown("Invalid response".to_string()));
+                return Err(Error::HttpError("Invalid response".to_string()));
```
Mubelotix, line 103: It may be the right time to remove these ugly things. What do you think?

Aeledfyr: Would it be reasonable to just store the `web-sys` error type?  I'm not sure if that would break anything - if that doesn't work, we could try converting it to a string and using the current system but with a better error message.

Mubelotix: The web-sys error type is awful and useless. It would be better to create an enum `WasmHttpError` or something like this.

Aeledfyr: I don't know enough about the potential http errors to be able to create a custom error enum for that.  For now we should either use the current system, pass the `web-sys` error on to the user, or store a stringified version of the web-sys error.  Implementing a custom error type for web errors seems like a bad idea, as it will likely miss some or many cases.

Aeledfyr:
This branch has been rebased off of the current master branch.  Besides the `wasm` `HttpError` discussion, this PR seems to be ready to merge.

curquiza:
If @Mubelotix is ok with that, let's merge it! Thanks for your involvement @Aeledfyr! 🙂

curquiza:
@Mubelotix and @Aeledfyr! If I'm not wrong these changes are breaking? I would say yes since `Http` is renamed into `HttpError` and `Unknown` error is removed.
(Not a big deal, I will update the minor if it indeed the case 🙂)

Mubelotix:
Yes, it is a breaking change and we *should* update the major version at the next release on crates.io

curquiza:
> Yes, it is a breaking change and we should update the minor version at the next release on crates.io

Yes, it will! Merging then!

curquiza:
Oups I squashed by habit (that's our process in the other SDKs repos)... sorry @Kerollmops! 😢

PR #40, opened 2020-07-31
Add code-sample file

Mubelotix:


Mubelotix:
I have added every samples. I just have to test them all and I will merge.

.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,494 @@
...
+  let movie_review_index: Index = client.get_index("movie_review").await.unwrap();
+  movie_review_index.update("movie_review_id").await.unwrap();
+delete_an_index_1: |-
+  movie_index.delete().await.unwrap();
+get_one_document_1: |-
+  let movie: Movie = movie_index.get_document(String::from("25684")).await.unwrap();
+get_documents_1: |-
+  let movie: Vec<Movie> = movie_index.get_documents(None, Some(2), None).await.unwrap();
+add_or_replace_documents_1: |-
+  // Note that the data differs a bit from other examples because we are using the exact same format as the database you downloaded, and the other examples are not
```
curquiza, line 22: No need to add this precision from my POV. What do you think @bidoubiwa?

.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,494 @@
...
+get_keys_1: |-
+  // unavailable for now
+get_settings_1: |-
+  let settings: Settings = movie_index.get_settings().await.unwrap();
+update_settings_1: |-
+  let mut synonyms = std::collections::HashMap::new();
+  synonyms.insert(String::from("wolverine"), vec![String::from("xmen"), String::from("logan")]);
+  synonyms.insert(String::from("logan"), vec![String::from("wolverine")]);
+
+  // That's quite ugly but we are working on an improvement
```
curquiza, line 78:
```suggestion
```

curquiza: 😇

.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,494 @@
...
+    "poster"
+  ];
+
+  let progress: Progress = movie_index.set_displayed_attributes(displayed_attributes).await.unwrap();
+reset_displayed_attributes_1: |-
+  let progress: Progress = movie_index.reset_displayed_attributes().await.unwrap();
+get_accept_new_fields_1: |-
+  let accept_new_fields: bool = movie_index.get_accept_new_fields().await.unwrap();
+update_accept_new_fields_1: |-
+  let progress: Progress = movie_index.set_accept_new_fields(false).await.unwrap();
```
curquiza, line 200:
```suggestion
```

curquiza: According to #42

.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,494 @@
...
+  // Build the query
+  // Warning: the testing database you downloaded does not contains the fields used in the filters so this example will not work on this database
+  let query: Query = Query::new("Batman")
+    .with_filters("release_date > 795484800 AND (director = \"Tim Burton\" OR director = \"Christopher Nolan\")");
+  
+  // Apply the query and get results
+  let results: SearchResults<Movie> = movie_index.search(&query).await.unwrap();
+filtering_guide_3: |-
+  // Build the query
+  // Warning: the testing database you downloaded does not contains the field "director" used in the filters so this example will not work on this database
```
curquiza, line 258: Same @bidoubiwa? What do you think about this comment?

.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,494 @@
...
+  // Build the query
+  let query: Query = Query::new("Avengers")
+    .with_filters("release_date > 795484800");
+  
+  // Apply the query and get results
+  let results: SearchResults<Movie> = movie_index.search(&query).await.unwrap();
+getting_started_create_index_md: |-
+  ```toml
+  [dependencies]
+  meilisearch-sdk = "0.2"
```
curquiza, line 395:
```suggestion
  meilisearch-sdk = "0.3.0"
```

curquiza: Plus, #44 makes us sure that the docs will always up-to-date with the last version

.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,494 @@
...
+    // here is where you have to insert any other sample of this documentation
+  }
+  ```
+
+  Please read the second example below before any other example of this documentation.
+
+  [About this crate](https://github.com/meilisearch/meilisearch-rust)
+getting_started_add_documents_md: |-
+  The Rust library is quite different from the other libraries since documents are stronly typed.
+  You have to implement the `Document` trait on a struct to be able to use it with Meilisearch.
```
curquiza, line 420: No need to detail that from my POV, since other SDKs are also typed (go and .NET for example) 😉

.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,494 @@
...
+    value: serde_json::Value,
+  }
+
+  impl Document for Movie {
+    type UIDType = String;
+    fn get_uid(&self) -> &Self::UIDType { &self.id }
+  }
+  ```
+
+  Then, you can add documents to an index like this:
```
curquiza, line 455:
```suggestion
  Then, add documents into the index:
```

.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,494 @@
...
+  let progress: Progress = movie_index.set_attributes_for_faceting(&["director", "producer", "genre", "production_companies"]).await.unwrap();
+faceted_search_walkthrough_facet_filters_1: |-
+  // Warning: the testing database you downloaded does not contains the fields used in the filters so this example will not work on this database
+  let query: Query = Query::new("thriller")
+    .with_facet_filters(vec![vec!["genres:Horror", "genres:Mystery"], vec!["director:Jordan Peele"]]);
+  let results: SearchResults<Movie> = movie_index.search(&query).await.unwrap();
+faceted_search_walkthrough_facets_distribution_1: |-
+  let query: Query = Query::new("Batman")
+    .with_facets_distribution(Some(vec!["genre"]));
+  let results: SearchResults<Movie> = movie_index.search(&query).await.unwrap();
```
curquiza, line 494: new line at the end of file 🙂

curquiza (changes requested):
Thanks for your HUGE work @Mubelotix! That's awesome!! 🚀

I commented some part and I also need the opinion of @bidoubiwa who is the champion of the documentation at Meili 😁

bidoubiwa (changes requested):

.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,494 @@
...
+      id: "287947".to_string(),
+      title: "Shazam".to_string(),
+      poster: "https://image.tmdb.org/t/p/w1280/xnopI5Xtky18MPhK40cZAGAOVeV.jpg".to_string(),
+      overview: "A boy is given the ability to become an adult superhero in times of need with a single magic word.".to_string(),
+      release_date: 1553299200,
+      genre: vec![
+        "Action".to_string(),
+        "Comedy".to_string(),
+        "Fantasy".to_string()
+      ]
```
bidoubiwa, line 34: We try to follow the context in which the code samples are showcased.
In this case it is [add or replace documents](https://docs.meilisearch.com/references/documents.html#add-or-replace-documents) from which their is a provided body above the code samples.

This body does not have the `genre` field.
```suggestion
```

.code-samples.meilisearch.yaml
bidoubiwa: It is not needed. Only the getting started is suppose to follow the movie dataset.
For every other example, it is important that it is clear in what it does. Clear enough that everything is understood just by reading the code.

.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,494 @@
...
+filtering_guide_1: |-
+  // Build the query
+  let query: Query = Query::new("Avengers")
+    .with_filters("release_date > 795484800");
+  
+  // Apply the query and get results
+  let results: SearchResults<Movie> = movie_index.search(&query).await.unwrap();
+filtering_guide_2: |-
+  // Build the query
+  // Warning: the testing database you downloaded does not contains the fields used in the filters so this example will not work on this database
```
bidoubiwa, line 250: These comment are not usefull. As only the getting started is following the movie dataset.
For every other example, the code should be enough to understand what is happening
```suggestion
```

.code-samples.meilisearch.yaml
bidoubiwa: I aggree with you, not needed.

.code-samples.meilisearch.yaml
```diff
@@ -0,0 +1,494 @@
...
+  // Warning: the testing database you downloaded does not contains the field "director" used in the filters so this example will not work on this database
+  let query: Query = Query::new("horror")
+    .with_filters("director = \"Jordan Peele\"");
+  
+  // Apply the query and get results
+  let results: SearchResults<Movie> = movie_index.search(&query).await.unwrap();
+filtering_guide_4: |-
+  
+  // Build the query
+  // Warning: the testing database you downloaded does not contains the fields used in the filters so this example will not work on this database
```
bidoubiwa, line 267:
```suggestion
```
Not needed.

.code-samples.meilisearch.yaml
bidoubiwa: I like the second sentence but the first is not needed.

bidoubiwa (changes requested):
Thanks a lot for the great work ! It is a real pleasure to have contributors doing such a fine work.

.code-samples.meilisearch.yaml
Mubelotix: Yes, but the patch version is not needed. You can write directly "0.3".

.code-samples.meilisearch.yaml
bidoubiwa: I understand you wanting to use the movie dataset for every code sample. But the idea in the documentation is to keep every example as small as possible.  By adding those unecessary lines it clutters the examples and make them less readable. 
We are going to add more examples for new features where you will need to add more fields if you want to keep that same dataset in the futur. This is not viable long term. It will make examples bigger and bigger.

Also, I'm working on changing some examples in the documentation with a different datasets ( see /meilisearch/documentation#493 ).

Lastly, I don't think a lot of rust users will be trying out every example of the documentation with your code. You had to because you made the code-samples, but users mostly go on the reference page of what the specifically needs and look at the example to understand the usage.

curquiza (approved):
Just awesome!!

curquiza:
bors try

curquiza:
Waiting for the @bidoubiwa's review and we will merge that 🚀

bidoubiwa (approved):

curquiza:
bors merge

Issue #29, opened 2020-07-29
Remove indexes after running the tests


curquiza:
When running the tests, we create indexes in the MeiliSearch instances, but we don't remove them once the tests have finished.
It leads to tests failure when I relaunch the tests with the same MeiliSearch instance, not really convenient for development purpose 🙂

irevoire:
Closed by #225

PR #46, opened 2020-08-21
Improve Query and SearchResults

### LINKED ISSUE #41: Improve Query and SearchResults
Query:
* Add the `matches` option
* Change `attributes_to_crop` into a Vec
* Change `attributes_to_highlight` into a Vec
* Change `attributes_to_retrieve` into a Vec

Results:
* Support matches_info
* Support formatted results

Mubelotix:
Will close #41 

### Query

- [x] Add the matches option
- [x] Change attributes_to_crop into an array
- [x] Change attributes_to_highlight into an array
- [x] Change attributes_to_retrieve into an array
- [x] Change method of the search route from GET to POST
- [x] Make the query string optional

### Results

- [x] Support matches_info
- [x] Support formatted results


Mubelotix:
Yes it's a breaking change so there will be some things to change in the sample file @curquiza

src/search.rs
```diff
@@ -1,17 +1,34 @@
...
+    pub formatted_result: Option<T>,
+    /// The object that contains information about the matches.
+    #[serde(rename = "_matchesInfo")]
+    pub matches_info: Option<HashMap<String, Vec<MatchRange>>>
+}
 
 #[derive(Deserialize, Debug)]
 #[serde(rename_all = "camelCase")]
 /// A struct containing search results and other information about the search.
 pub struct SearchResults<T> {
```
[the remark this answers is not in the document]
Mubelotix: I am not sure if it would be useful in some cases. But is this possible without making a new method ? If yes we could do it anyway

src/search.rs
```diff
@@ -111,31 +154,31 @@ impl<'a> Query<'a> {
             ..self
         }
     }
-    pub fn with_facet_filters(self, facet_filters: Vec<Vec<&'a str>>) -> Query<'a> {
+    pub fn with_facet_filters(self, facet_filters: &'a[&'a[&'a str]]) -> Query<'a> {
```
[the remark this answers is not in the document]
Mubelotix: Thanks I learned something ! I will add a commit (but not before the end of the week)

src/search.rs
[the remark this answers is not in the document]
Kerollmops: I'm not sure I understand, why would it need a new method? The `T` type must implement `Ser/Deserialize` so it works with `serde_json::Value` if the user doesn't specify the `T` type itself.

Ask me any question, I am not sure my message above helps much, and also, this is not mandatory to implement.

Mubelotix:
@curquiza I think it would be better to write new tests when we improve the existing ones (see #29). There will be major changes soon.

curquiza:
Okay, @Mubelotix, but it should be done soon in case there is an error. For the momebt, I have no proof it works without any test.

curquiza:
Have all the changes asked @Kerollmops been taken into account? Is this PR ready for merging?

Mubelotix:
Let's merge fast so we can merge #40 too.

README.md
```diff
@@ -111,7 +111,7 @@ async fn main() {
     ], Some("book_id")).await.unwrap();
 
     // Query books (note that there is a typo)
-    let query = Query::new("harry pottre");
+    let query = Query::new().with_query("harry pottre").build();
     println!("{:?}", books.search::<Book>(&query).await.unwrap().hits);
```
Kerollmops, line 115: Hey, I maybe have a better idea for this builder, maybe more something like that. What do you think?

```rust
books.search::<Book>().with_query("harry pottre").build().await?;
```

src/search.rs
```diff
@@ -1,202 +1,261 @@
...
+            offset: None,
+            limit: None,
+            filters: None,
+            facet_filters: None,
+            facets_distribution: None,
+            attributes_to_retrieve: None,
+            attributes_to_crop: None,
+            crop_length: None,
+            attributes_to_highlight: None,
+            matches: None,
```
Kerollmops, line 206: Maybe you can just call `Query::default()` here?

src/search.rs
```diff
@@ -1,202 +1,261 @@
...
 
+fn serialize_with_wildcard<S, T>(data: &Option<Selectors<T>>, s: S) -> Result<S::Ok, S::Error> where S: Serializer, T: Serialize {
+    match data {
+        Some(Selectors::All) => ["*"].serialize(s),
+        Some(Selectors::Some(data)) => data.serialize(s),
+        None => s.serialize_none(),
+    }
+}
+
+fn serialize_attributes_to_crop_with_wildcard<S>(data: &Option<Selectors<&[AttributeToCrop]>>, s: S) -> Result<S::Ok, S::Error> where S: Serializer {
```
Kerollmops, line 58:
```suggestion
fn serialize_attributes_to_crop_with_wildcard<S: Serializer>(data: &Option<Selectors<&[AttributeToCrop]>>, s: S) -> Result<S::Ok, S::Error> {
```

src/search.rs
```diff
@@ -1,202 +1,261 @@
...
     pub exhaustive_facets_count: Option<bool>,
-    /// processing time of the query
+    /// Processing time of the query
     pub processing_time_ms: usize,
-    /// query originating the response
+    /// Query originating the response
     pub query: String,
 }
 
+fn serialize_with_wildcard<S, T>(data: &Option<Selectors<T>>, s: S) -> Result<S::Ok, S::Error> where S: Serializer, T: Serialize {
```
Kerollmops, line 50:
```suggestion
fn serialize_with_wildcard<S: Serializer, T: Serialize>(data: &Option<Selectors<T>>, s: S) -> Result<S::Ok, S::Error> {
```

src/search.rs
```diff
@@ -1,202 +1,261 @@
...
+        self.crop_length = Some(crop_length);
+        self
+    }
+    pub fn with_matches<'b>(&'b mut self, matches: bool) -> &'b mut Query<'a> {
+        self.matches = Some(matches);
+        self
+    }
+    pub fn build(&mut self) -> Query<'a> {
+        self.clone()
     }
```
Kerollmops, line 255: Maybe it would be better to consume the query itself `fn build(self) -> Query<'a> { self }` instead. This way we avoid cloning every time we query and if the user wants to do multiple queries with the same parameters he can clone the builder before building it.

Kerollmops (changes requested):
This is great work 😄 But maybe you could chnage the builder a little bit!

README.md
Mubelotix: That's a great idea! It will be done. But I think we should keep a way to build a `Query` and apply it after.

src/search.rs
Mubelotix: The problem here is that when we use the builder syntax, we return `&mut Self` at each method. So the build method, which is used at the end of the chain, cannot take `Self`. I think cloning is not a problem since it contains only options and references. Cloning is extremely cheap.
But what if we implement the `Future` trait and remove the build method?

Kerollmops: Yeah, you are right, this is a near free clone :)

curquiza:
bors try

curquiza:
@Mubelotix, the tests fail when you make your branch up to date with master. It would be impossible to merge even if Kero approves your PR.
Maybe you should [rebase your branch](https://gist.github.com/curquiza/5f7ce615f85331f083cd467fc4e19398) to check what's wrong 🙂

Kerollmops (approved):
Thank you very much for these changes, I started to complain about the `build`, `execute` and `execute_query` but it is not that bad in fact.

I think the `Index::execute_query` is here just to be able to use the same query on multiple indexes 👍

curquiza:
bors try

curquiza (approved):
🚀

curquiza:
bors merge

Issue #48, opened 2020-09-24
Don't require tokio runtime


MCOfficer:
Currently, this project requires usage of the tokio runtime, using it with f.e. async-std is not possible. It would be nice if we could also use it with these other runtimes, without having to do this (yuck):
```rust
    tokio::runtime::Builder::new()
        .basic_scheduler()
        .enable_all()
        .build()?
        .block_on(do_something_async())?;
```

As far as i can see, the only reason is the dependency of reqwest.

Switching to a runtime-agnostic http client seems possible. I made out https://github.com/http-rs/surf as a good option, especially since it has a wasm-backend. One should probably wait for its [2.0 release](https://github.com/http-rs/surf/milestone/1) before switching.

Mubelotix:
Do without `reqwest` would indeed be great. I never heard about `surf`. Is it as stable as `reqwest`?

MCOfficer:
Once it reaches 2.0, I'd expect it to be quite stable.

Firstly, the http-rs is one of the bigger players in the whole networking topic, they know what they're doing. Secondly, surf is just an abstraction over two very stable http backends - `hyper` (which reqwest uses) and `isahc` (a curl wrapper). 

As it stands, #49 is using the `isahc`, but switching is as simple as adding a feature.

MCOfficer:
I read up on it, and unless you are paranoid surf, is perfectly stable.

If you are paranoid, [surf is just as unstable as reqwest](https://medium.com/@shnatsel/smoke-testing-rust-http-clients-b8f2ee5db4e6), not matter which backend it uses. But at that point, *nothing* is stable.

Mubelotix:
Since I have read [this article](https://medium.com/@shnatsel/smoke-testing-rust-http-clients-b8f2ee5db4e6) I am quite paranoid about the stability. But you should be right.

curquiza:
Hello @MCOfficer! Welcome here, and thanks for your suggestion 🙂

@Kerollmops what do you think about `reqwest` vs `surf`?

(FYI @MCOfficer, @Kerollmops is the main contributor of the [MeiliSearch's core engine](https://github.com/meilisearch/MeiliSearch), so written in Rust)

Kerollmops:
Hey @MCOfficer,

[...]

PR #49, opened 2020-09-24
Use isahc as replacement for reqwest

MCOfficer:
This PR adresses #48 by using surf instead of reqwest.

In #48 I said that surf could be used from wasm as well - turns out it's not quite there yet: https://github.com/http-rs/surf/issues/210

Marking WIP because you probably don't want to merge this before surf 2.0.

---

Switched to isahc instead of surf, see comments.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ serde = { version = "1.0", features = ["derive"] }
 urlencoding = "1.0"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-reqwest = "0.10"
+isahc = "0.9"
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 web-sys = { version = "0.3", features = ["RequestInit", "Headers", "Window", "Response", "console"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,12 +15,16 @@ serde = { version = "1.0", features = ["derive"] }
 urlencoding = "1.0"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
-isahc = "0.9"
+isahc = { version = "0.9", features = ["http2", "text-decoding"], default_features = false }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 web-sys = { version = "0.3", features = ["RequestInit", "Headers", "Window", "Response", "console"] }
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
+
+[features]
+default = ["isahc-static-curl"]
+isahc-static-curl = ["isahc/static-curl"]
 
 [dev-dependencies]
 env_logger = "0.7"
--- a/src/request.rs
+++ b/src/request.rs
@@ -25,29 +25,29 @@ pub(crate) async fn request<Input: Serialize + std::fmt::Debug, Output: 'static 
         Method::Get => Request::get(url)
             .header("X-Meili-API-Key", apikey)
             .body(())
-            .map_err(|e| isahc::Error::InvalidHttpFormat(e))?
+            .map_err(isahc::Error::InvalidHttpFormat)?
             .send_async().await?,
         Method::Delete => Request::delete(url)
             .header("X-Meili-API-Key", apikey)
             .body(())
-            .map_err(|e| isahc::Error::InvalidHttpFormat(e))?
+            .map_err(isahc::Error::InvalidHttpFormat)?
             .send_async().await?,
         Method::Post(body) => Request::post(url)
             .header("X-Meili-API-Key", apikey)
             .header("Content-Type", "application/json")
             .body(to_string(&body).unwrap())
-            .map_err(|e| isahc::Error::InvalidHttpFormat(e))?
+            .map_err(isahc::Error::InvalidHttpFormat)?
             .send_async().await?,
         Method::Put(body) => Request::put(url)
             .header("X-Meili-API-Key", apikey)
             .header("Content-Type", "application/json")
             .body(to_string(&body).unwrap())
-            .map_err(|e| isahc::Error::InvalidHttpFormat(e))?
+            .map_err(isahc::Error::InvalidHttpFormat)?
             .send_async().await?,
     };
 
     let status = response.status().as_u16();
-    let mut body = response.text_async().await.map_err(|e| isahc::Error::Io(e))?;
+    let mut body = response.text_async().await.map_err(isahc::Error::Io)?;
     if body.is_empty() {
         body = "null".to_string();
     }
--- a/src/request.rs
+++ b/src/request.rs
@@ -2,7 +2,6 @@ use crate::errors::Error;
 use log::{error, trace, warn};
 use serde::{de::DeserializeOwned, Serialize};
 use serde_json::{from_str, to_string};
-use isahc::prelude::*;
 
 #[derive(Debug)]
 pub(crate) enum Method<T: Serialize> {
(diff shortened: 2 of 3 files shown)


Mubelotix:
Even if we switch to `surf`, I think we don't need it on WASM. Currently, we are using web-sys to access web APIs. It is quite simple and it works great. Maybe we could try both solutions and choose the one producing the smaller wasm file.

curquiza:
Hello @MCOfficer, thanks for drafting this PR.
The test are failing because of my mistake. It's fixed now, so you can [rebase your branch](https://gist.github.com/curquiza/5f7ce615f85331f083cd467fc4e19398) and it should be ok now. Sorry for that!

MCOfficer:
Will rebase eventually.

For now, I'm worried about surf. It seems to time out a lot when requests are run in parallel:
image


I also had this issue in a private project (in fact, the "got 200 for X" requests you see in the log used to be done by surf). It kept timing out perfectly healthy websites. Using isahc instead works perfectly fine, as you can see.

What's your take on using isahc directly? the only alternative would be surf's other backend, `async-h1`. I didn't choose it because i want more than just HTTP1.1, but for Meilisearch it should be fine.

Mubelotix:
The timeout problem occurs with `reqwest` too. It's a reason why we can't run tests in parallel on this repo. It seems to be a server-side issue, but since it uses `actix-web`, it should not occur. I also dislike the idea of using HTTP1.1. It would be slower. We can try isahc and see if it is suitable.

MCOfficer:
> The timeout problem occurs with `reqwest` too. It's a reason why we can't run tests in parallel on this repo. It seems to be a server-side issue, but since it uses `actix-web`, it should not occur. I also dislike the idea of using HTTP1.1. It would be slower. We can try isahc and see if it is suitable.

Missing information: in my tests the timeouts occured any website, nothing to do with meilisearch.

src/request.rs
```diff
@@ -20,25 +21,33 @@ pub(crate) async fn request<Input: Serialize + std::fmt::Debug, Output: 'static
...
-
-    let response = match &method {
-        Method::Get => client.get(url).header("X-Meili-API-Key", apikey).send().await?,
-        Method::Delete => client.delete(url).header("X-Meili-API-Key", apikey).send().await?,
-        Method::Post(body) => client.post(url)
+    let mut response = match &method {
+        Method::Get => Request::get(url)
+            .header("X-Meili-API-Key", apikey)
+            .body(())
+            .map_err(|e| isahc::Error::InvalidHttpFormat(e))?
```
MCOfficer, line 28: this is a bit ugly, should I add an `impl From<isahc::http::Error> for Error` instead?

MCOfficer (commented):
Squashed and rebased while switching to isahc.

Here's a few comments. Left to do: add a passthrough flag for isahc's `static-curl` flag.

src/request.rs
Mubelotix: It only takes one line. It's not useful to implement `From<isahc::http::Error>`.

MCOfficer: 4x1 line, about as much as the `From<isahc::Error>`. But i have no strong opinion here.

Cargo.toml
```diff
@@ -15,13 +15,17 @@ serde = { version = "1.0", features = ["derive"] }
...
+isahc = { version = "0.9", features = ["http2", "text-decoding"], default_features = false }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 web-sys = { version = "0.3", features = ["RequestInit", "Headers", "Window", "Response", "console"] }
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
 
+[features]
+default = ["isahc-static-curl"]
+isahc-static-curl = ["isahc/static-curl"]
```
Mubelotix, line 27: This feature must not be available for wasm targets.

src/request.rs
```diff
@@ -2,6 +2,7 @@ use crate::errors::Error;
 use log::{error, trace, warn};
 use serde::{de::DeserializeOwned, Serialize};
 use serde_json::{from_str, to_string};
+use isahc::prelude::*;
```
Mubelotix, line 5: Unavailable for wasm builds. Move this into the function.

Mubelotix (changes requested):
It seems that `isahc` is more performant and leads to smaller build binaries.
There are several problems with this PR on wasm targets. You also have to check the code with `cargo check --target=wasm32-unknown-unknown`.
There is also a huge amount of small modifications required in the documentation. We have to remove `tokio`.

Cargo.toml
MCOfficer: Restricting features to certain targets is not (yet) possible in stable: https://github.com/rust-lang/cargo/issues/1197

Besides, without `isahc` being compiled, this feature won't do anything.

MCOfficer:
> There is also a huge amount of small modifications required in the documentation. We have to remove `tokio`.

Do you want me to do this (or at least attempt to)?

Mubelotix:
> > There is also a huge amount of small modifications required in the documentation. We have to remove `tokio`.
>
> Do you want me to do this (or at least attempt to)?

That would be great but I can do it.

Mubelotix:
@curquiza What is the best way for me to add documentation? Merging to master and making a second PR quickly? Or merging to another branch so we then make a PR gathering all the changes?

MCOfficer:
> > > There is also a huge amount of small modifications required in the documentation. We have to remove `tokio`.
> > 
> > 
> > Do you want me to do this (or at least attempt to)?
> 
> That would be great but I can do it.

Sorry, i'm kinda busy atm. If you have time, feel free to do it in my stead. thanks :)

curquiza:
Hello @MCOfficer and @Mubelotix!
I've just created a new branch `isahc-dev`. We will merge this PR first into `isahc-dev`. Then, when you have time @Mubelotix, you can do as many PRs as you need (pointing to `isahc-dev`) to complete the @MCOfficer work 🙂
No rush with this solution! We don't like rushing at Meili.

Is this PR ready to be merged? I see the tests are failing. Do we merge anyway?

MCOfficer:
> Is this PR ready to be merged? I see the tests are failing. Do we merge anyway?

It was ready when I checked a few days ago, assuming @Mubelotix doesn't veto. Regarding the failing pipeline, I can't reproduce that. The tests were working, but the pipeline straight up fails to compile.

curquiza:
Ok, I cannot merge to `master` if the CI is failing, but I can accept the merge into `isahc-dev` and then we will figure out why it fails and will fix it 🤞
@Mubelotix if you are ok with this PR, feel free to merge it into `isahc-dev`. I already changed the "merge destination" so you just have to click on the merge button on the GitHub interface

Capture d’écran 2020-11-20 à 10 13 06

Mubelotix:
CI is failing due to tokio. Since the next step is to remove tokio, we can ignore this. Do you want me to do another PR to `isahc-dev` or should I push commits directly to `isahc-dev` and then we merge to `master`?

Mubelotix (approved):

MCOfficer:
Then you take it from here. Thank you for working on meilisearch, and best of luck going forward!

curquiza:
> Do you want me to do another PR to isahc-dev or should I push commits directly to isahc-dev and then we merge to master?

@Mubelotix, if possible, create another PR(s), but don't push directly to `isahc-dev` -> it would be easier to review 👍

MCOfficer:
Ok, this took me completely by surprise. Thank you <3



image

curquiza:
So coool! 🚀  And sorry for the delay, we had troubles with some contributor letters like yours! 🥲

PR #89, opened 2021-01-13
Note to mention that it is NOT running in browser

rrjanbiah:
Added a note to mention that the tutorial is not about running MeiliSearch in browser using WASM

### DIFF
--- a/examples/web_app/README.md
+++ b/examples/web_app/README.md
@@ -1,5 +1,8 @@
 # Build your front-end page in Rust with WebAssembly
 
+> **Note**
+> If you're looking to run MeiliSearch in a web browser using WASM, that is not possible yet as the Rust libraries used by MeiliSearch are not supporting WASM. Instead, this tutorial explains how to embed the whole frontend code in a WASM.
+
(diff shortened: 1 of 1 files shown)


examples/web_app/README.md
```diff
@@ -1,5 +1,8 @@
 # Build your front-end page in Rust with WebAssembly
 
+> **Note**
+> If you're looking to run MeiliSearch in a web browser using WASM, that is not possible yet as the Rust libraries used by MeiliSearch are not supporting WASM. Instead, this tutorial explains how to embed the whole frontend code in a WASM.
```
Mubelotix, line 4:
```suggestion
> It is not possible to run MeiliSearch in the browser without a server. This example will only make use of the API.
```

Mubelotix (changes requested):

examples/web_app/README.md
curquiza:
```suggestion
> It is not possible to run MeiliSearch in the browser without a server. This demo uses the Rust SDK in a browser using WASM, and communicate with a MeiliSearch instance that is running on a remote server.
```

PR #108, opened 2021-02-15
Adding get_all_updates route

### LINKED ISSUE #39: Missing routes
The `Get Keys` and `Get all update status` routes are still missing. It can be useful in rare cases.

sanders41:
Resolves #39

The reason for renaming some of the indexes in the docs was because with them having the same names they were having side effects on each other and causing the doc tests to fail. Is this renaming acceptable, or is there a better way to handle this? The problem is related to issue #29.

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -62,7 +62,7 @@ search_1: |-
 get_update_1: |-
   let status: Status = progress.get_status().await.unwrap();
 get_all_updates_1: |-
-  // unavailable for now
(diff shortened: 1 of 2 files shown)


curquiza:
Hello @sanders41!

Thanks for this PR!
The re-naming is the best temporary choice: tests should indeed be independent. I've just opened an issue about it #109.

curquiza (approved):
LGTM 👌 

Thanks again 😁

curquiza:
bors merge

Issue #109, opened 2021-02-15
Make each test independent


curquiza:
To avoid side effects between each test, the tests should all be independent.

It means, at the beginning of each test:
- create the needed index(es)
- push documents according to the test
- update settings if necessary

And at the end of each test:
- remove the index(es)

This is linked to #29

sanders41:
I have looked into creating and deleting the index with each test and the problem I have run into is because the tests are running in threads, the tests still fail. The reason for this is multiple tests can be running at the same time trying to use the same index name. One solution to this is to run the tests all in a single thread with `cargo test -- --test-threads=1`. By doing this the tests no longer overlap and will pass, but the test take MUCH longer to run.

Mubelotix:
@sanders41 I have already thought about this issue. Your third solution has already been applied [here](https://github.com/meilisearch/meilisearch-rust/blob/a78ec535dab6e3fdbb34a9d72900aef25fa289dc/src/search.rs#L284) and I plan to extend this structure to all other tests. However, there is another issue that you did not expect. For a reason I am unable to explain when tests make simultaneous requests to the MeiliSearch server, requests will hang and never get answers. So we cannot remove `test-threads=1` anyway.

sanders41:
> For a reason I am unable to explain when tests make simultaneous requests to the MeiliSearch server, requests will hang and never get answers.

I have seen this issue also. I was guessing it was happening because of the index overlap, but sounds like your testing shows that isn’t the cause.

irevoire:
Closed by #225

PR #76, opened 2020-12-09
Update old code

Mubelotix:
I copied a few data structures from the meilisearch core project when I started this SDK.
It seems that these structs and enums were modified during the last months and that breaking changes were introduced.
It would be useful if GitHub allowed me to subscribe to the changes on the 2 concerned files but it's not possible.

### TODO

- [x] Update the outdated code
- [x] Add new tests so that we can detect breaking changes faster (note that test structure is not ideal and there will be a new PR about this)

PR #110, opened 2021-02-15
Making duration and processed_at optional

### LINKED ISSUE #107: Error with ProcessedStatus
I was looking at the possibility of adding the get all update status route in #39 and have run into an issue. According to the [documentation](https://docs.meilisearch.com/reference/api/updates.html#get-all-update-status) the response will look like:

```json
[
  {
    "status": "processed",
    "updateId": 1,
    "type": {
      "name": "DocumentsAddition",
      "number": 4
    },
    "duration": 0.076980613,
    "enqueuedAt": "2019-12-07T21:16:09.623944Z",
    "processedAt": "2019-12-07T21:16:09.703509Z"
  }
]

```

but when I test I get an error ```ParseError(Error("missing field `duration`", line: 1, column: 325)```. This error seems to resolve and the response does include the duration and processedAt information if I do `sleep(Duration::from_secs(1));` after adding documents.

I checked what response I get in the [python clients test](https://github.com/meilisearch/meilisearch-python/blob/628552573181c4d348489362e694d3d2a32372ab/meilisearch/tests/index/test_index_update_meilisearch.py#L11) and the result in the test is:

```py
[
    {
        'status': 'enqueued', 
        'updateId': 0,
        'type': {
            'name': 'DocumentsAddition',
            'number': 30
        },
        'enqueuedAt': '2021-02-14T14:07:09.364505700Z'
    },
    {
        'status': 'enqueued',
        'updateId': 1,
        'type': {
             'name': 'DocumentsAddition',
             'number': 30
        },
        'enqueuedAt': '2021-02-14T14:07:09.373833600Z'
    }
]
```

Since Python is putting the information into a dictionary the tests passes, but if the end user is expecting the keys to be there and used, for example, `result["duration"]` instead of `result.get("duration")` this would error.

My guess is that in the documentation it is assuming that the documents will always be processed before checking the status, but this doesn't seem to be the case? If that is correct would updating [ProcessedStatus](https://github.com/meilisearch/meilisearch-rust/blob/ae1df1208bc8a6647905b9758705687ea4ca4839/src/progress.rs#L111) to the following be appropriate?

```rs
#[derive(Deserialize, Debug)]
pub struct ProcessedStatus {
    pub update_id: u64,
    #[serde(rename = "type")]
    pub update_type: UpdateType,
    pub error: Option<String>,
    pub duration: Option<f64>,        // in seconds
    pub enqueued_at: String,  // TODO deserialize to datatime
    pub processed_at: Option<String>, // TODO deserialize to datatime
}
```

Also, if this is the case should the documentation be updated to note that these fields are optional in the response since it would affect all clients even if they don't error?

sanders41:
Resolves  #107

Note that this is a breaking change because duration and processed_at will now need to either be matched to Some or unwrapped.

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -114,9 +114,11 @@ pub struct ProcessedStatus {
     pub update_type: UpdateType,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub error: Option<String>,
-    pub duration: f64,        // in seconds
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub duration: Option<f64>,        // in seconds
     pub enqueued_at: String,  // TODO deserialize to datatime
-    pub processed_at: String, // TODO deserialize to datatime
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub processed_at: Option<String>, // TODO deserialize to datatime
 }
 
 #[derive(Debug, Clone, Deserialize)]


sanders41:
Before merging this I have a question. Maybe ProcessedStatus is fine as it is because I now see that there is both a ProcessedStatus and Enqueued that are separate. After reading https://github.com/meilisearch/meilisearch-rust/issues/109 I think I had a misunderstanding initially and that the response would always be processed. If that is the case instead of this update then maybe there should be an update to [get_all_updates](https://github.com/meilisearch/meilisearch-rust/blob/208c138cee1943f1701bfc6ec5cfea88825aba08/src/indexes.rs#L662) to return either ProcessedStatus or Enqueued status depending on what it receives?

sanders41:
After looking at this further with my new understanding the better option seems to be to:

1. Close this PR
2. Open a new PR to update [get_all_updates](https://github.com/meilisearch/meilisearch-rust/blob/208c138cee1943f1701bfc6ec5cfea88825aba08/src/indexes.rs#L662) to:

```rust
pub async fn get_all_updates(&self) -> Result<Vec<Status>, Error> {
        let values = request::<(), Vec<Value>>(
            &format!(
                "{}/indexes/{}/updates",
                self.client.host, self.uid
            ),
            self.client.apikey,
            Method::Get,
            200,
        )
        .await?;

        let mut status_values: Vec<Status> = Vec::new();

        for value in values {
            if let Ok(status) = from_value::<ProcessedStatus>(value.clone()) {
                status_values.push(Status::Processed(status));
            } else {
                let result = from_value::<EnqueuedStatus>(value);
                match result {
                    Ok(status) => status_values.push(Status::Enqueued(status)),
                    Err(e) => return Err(Error::ParseError(e)),
                }
            }
        }

        Ok(status_values)
    }
```

Thoughts?

Mubelotix:
Firstly, you might want to look at this #76 since you are actually working on deprecated structs that will be changed in this PR.
It seems at the first view that the bug of #107 is still present [here](https://github.com/meilisearch/meilisearch-rust/blob/d610bc6805bdc243a9d9a1b92360428071369c4b/src/progress.rs#L112).
But your last comments are right. The API is never returning `ProcessedStatus` (`ProcessedUpdateResult` in #76) directly. Instead, it is returning `Status` (`UpdateStatus` in #76) which is a serde-tagged enum. Maybe #107 should be fixed in #76 too? We plan to merge #76 before the end of the week.

sanders41:
Thanks @Mubelotix. Looking through #76 it looks like once that is merged get_all_updates should return the UpdateStatus enum meaning the above suggested update is the right direction, but would need to also check for a Failed status?

If that is correct will the Failed status have all the information present for ProcessedUpdateResult including duration and processed_at? If yes and returning UpdateStatus is correct for get_all_updates #107 would be resolved.

Mubelotix:
> will the Failed status have all the information present for ProcessedUpdateResult including duration and processed_at?

Yes. The meilisearch server considers that a failed task is still a processed task and will contain the exact same information.

> returning UpdateStatus is correct for get_all_updates #107 would be resolved

I think so ! Let us merge #76 and we will be able to fix that.

sanders41:
Sounds good. I have the update ready and now just need to add the Failed option. I'll work on adding that in so it's ready to go when #76 gets merged.

sanders41:
@curquiza here is where I think we are with this if you agree.

With some initial testing last night it looks like #76 simplifies the update needed for get_all_updates and it would now become:

```rust
pub async fn get_all_updates(&self) -> Result<Vec<UpdateStatus>, Error> {
    request::<(), Vec<UpdateStatus>>(
        &format!(
            "{}/indexes/{}/updates",
            self.client.host, self.uid
        ),
        self.client.apikey,
        Method::Get,
        200,
    )
    .await
}
```

Then after #76 and the above update are made #107 is not longer an issue making this PR unnecessary so I will close it.

curquiza:
Thanks @sanders41 for the update!
Sorry you spent time on it!

sanders41:
No worries, I learned more about the way the client works in the process so it was worth the time :smile:

PR #111, opened 2021-02-20
Adding the wait_for_pending_update method

### LINKED ISSUE #77: waitForPendingUpdate method
The `wait_for_pending_update` method is a method to wait for the asynchronous update of MeiliSearch (like update/reset the settings or add/update/delete documents).
This method is really useful for testing phases: you add documents, you wait for the document update to finish, you check all the documents have been added. This is also used by users who wanted to work synchronously.

## Details of the method

- This method is a loop that checks the status of the update every `interval` until a `timeout`.
Both parameters (`interval` and `timeout`) must be:
  - customizable by the user.
  - optional and the method should set default values if they are not defined by the user. 
    `interval`: 50ms
    `timeout`: 5 seconds

- If the method times out, it should return a TimeOut Exception.

## Example

In the PHP SDK
https://github.com/meilisearch/meilisearch-php/blob/b6e588d739f78071d9361d4eea9b05764f22877b/src/Endpoints/Indexes.php#L104-L116

## Steps

- add in the code
- tests

sanders41:
Resolves  #77

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -15,8 +15,6 @@ pub enum Error {
         error_link: String,
     },
 
-    /// The MeiliSearch request timed out.
-    MeiliSearchTimeoutError { message: String },
     /// There is no MeiliSearch server listening on the [specified host]
     /// (../client/struct.Client.html#method.new).
     UnreachableServer,
@@ -259,7 +257,6 @@ impl std::fmt::Display for Error {
                 message,
                 error_link,
             ),
-            Error::MeiliSearchTimeoutError { message } => write!(fmt, "MeiliSearch {}", message),
             Error::UnreachableServer => write!(fmt, "The MeiliSearch server can't be reached."),
             Error::InvalidRequest => write!(fmt, "Unable to generate a valid HTTP request. It probably comes from an invalid API key."),
             Error::ParseError(e) => write!(fmt, "Error parsing response JSON: {}", e),
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -15,6 +15,8 @@ pub enum Error {
         error_link: String,
     },
 
+    /// The MeiliSearch request timed out.
+    MeiliSearchTimeoutError { message: String },
     /// There is no MeiliSearch server listening on the [specified host]
     /// (../client/struct.Client.html#method.new).
     UnreachableServer,
@@ -257,6 +259,7 @@ impl std::fmt::Display for Error {
                 message,
                 error_link,
             ),
+            Error::MeiliSearchTimeoutError { message } => write!(fmt, "MeiliSearch {}", message),
             Error::UnreachableServer => write!(fmt, "The MeiliSearch server can't be reached."),
             Error::InvalidRequest => write!(fmt, "Unable to generate a valid HTTP request. It probably comes from an invalid API key."),
             Error::ParseError(e) => write!(fmt, "Error parsing response JSON: {}", e),
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -56,6 +56,9 @@ impl<'a> Progress<'a> {
 
     /// Wait until MeiliSearch processes an update, and get its status.
     /// 
+    /// interval_ms = The frequency at which the server should be polled in milliseconds. Default = 50ms
+    /// timeout_ms = The maximum time to wait for processing to complete in milliseconds. Default = 5000ms
+    /// 
     /// # Example
     ///
     /// ```
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -3,6 +3,8 @@
 use crate::{errors::Error, indexes::Index, request::*};
 use serde::Deserialize;
 use std::collections::{BTreeMap, BTreeSet};
+use std::thread::sleep;
+use std::time::Duration;
 
 #[derive(Deserialize)]
 #[serde(rename_all = "camelCase")]
@@ -36,6 +38,7 @@ impl<'a> Progress<'a> {
     /// let mut movies_index = client.get_or_create("movies").await.unwrap();
     /// let progress = movies_index.delete_all_documents().await.unwrap();
     /// let status = progress.get_status().await.unwrap();
+    /// # client.delete_index("movies").await.unwrap();
     /// # });
     /// ```
     pub async fn get_status(&self) -> Result<UpdateStatus, Error> {
@@ -49,6 +52,89 @@ impl<'a> Progress<'a> {
             200,
         )
         .await
+    }
+
+    /// Wait until MeiliSearch processes an update, and get its status.
+    /// 
+    /// # Example
+    ///
+    /// ```
+    /// # use meilisearch_sdk::{client::*, document, indexes::*, progress::*};
+    /// # use serde::{Serialize, Deserialize};
+    /// 
+    /// # #[derive(Debug, Serialize, Deserialize, PartialEq)]
+    /// # struct Document {
+    /// #    id: usize,
+    /// #    value: String,
+    /// #    kind: String,
+    /// # }
+    /// # 
+    /// # impl document::Document for Document {
+    /// #    type UIDType = usize;
+    /// #
+    /// #    fn get_uid(&self) -> &Self::UIDType {
+    /// #        &self.id
+    /// #    }
+    /// # }
+    /// 
+    /// # futures::executor::block_on(async move {
+    /// let client = Client::new("http://localhost:7700", "masterKey");
+    /// let movies = client.create_index("movies_wait_for_pending", None).await.unwrap();
+    /// 
+    /// let progress = movies.add_documents(&[
+    ///     Document { id: 0, kind: "title".into(), value: "The Social Network".to_string() },
+    ///     Document { id: 1, kind: "title".into(), value: "Harry Potter and the Sorcerer's Stone".to_string() },
+    /// ], None).await.unwrap();
+    /// 
+    /// let status = progress.wait_for_pending_update(None, None).await.unwrap();
+    /// 
+    /// # client.delete_index("movies_wait_for_pending").await.unwrap();
+    /// assert!(matches!(status, UpdateStatus::Processed { .. }));
+    /// # });
+    /// ```
+    pub async fn wait_for_pending_update(
+        &self,
+        interval_ms: Option<u64>,
+        timeout_ms: Option<u64>,
+    ) -> Result<UpdateStatus, Error> {
+        let interval: u64;
+        let timeout: u64;
+
+        match interval_ms {
+            Some(v) => interval = v,
+            None => interval = 50,
+        }
+
+        match timeout_ms {
+            Some(v) => timeout = v,
+            None => timeout = 5000,
+        }
+
+        let mut elapsed_time = 0;
+        let mut status: UpdateStatus;
+
+        while timeout > elapsed_time {
+            status = self.get_status().await?;
+
+            match status {
+                UpdateStatus::Failed { .. } | UpdateStatus::Processed { .. } => {
+                    return self.get_status().await;
+                },
+                UpdateStatus::Enqueued { .. } => {
+                    elapsed_time += interval;
+                    sleep(Duration::from_millis(interval));
+                },
+            };
+        }
+
+        Err(
+            Error::MeiliSearchTimeoutError {
+                message: format!(
+                    "timeout of {}ms has been exceeded when waiting for pending update to resolve.",
+                    timeout
+                ),
+            }
+        )
     }
 }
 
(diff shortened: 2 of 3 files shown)


src/errors.rs
```diff
@@ -15,6 +15,8 @@ pub enum Error {
         error_link: String,
     },
 
+    /// The MeiliSearch request timed out.
+    MeiliSearchTimeoutError { message: String },
```
Mubelotix, line 19: I don't think it's a good idea to add a new variant in this case since there is only one constructor on this variant.

src/progress.rs
```diff
@@ -50,6 +53,92 @@ impl<'a> Progress<'a> {
...
+    /// 
+    /// interval_ms = The frequency at which the server should be polled in milliseconds. Default = 50ms
+    /// timeout_ms = The maximum time to wait for processing to complete in milliseconds. Default = 5000ms
+    /// 
+    /// # Example
+    ///
+    /// ```
+    /// # use meilisearch_sdk::{client::*, document, indexes::*, progress::*};
+    /// # use serde::{Serialize, Deserialize};
+    ///
```
Mubelotix, line 67:
```suggestion
    /// #
```

src/progress.rs
```diff
@@ -50,6 +53,92 @@ impl<'a> Progress<'a> {
...
+    /// # }
+    /// # 
+    /// # impl document::Document for Document {
+    /// #    type UIDType = usize;
+    /// #
+    /// #    fn get_uid(&self) -> &Self::UIDType {
+    /// #        &self.id
+    /// #    }
+    /// # }
+    ///
```
Mubelotix, line 82:
```suggestion
    /// #
```

src/progress.rs
```diff
@@ -50,6 +53,92 @@ impl<'a> Progress<'a> {
...
+    /// 
+    /// let status = progress.wait_for_pending_update(None, None).await.unwrap();
+    /// 
+    /// # client.delete_index("movies_wait_for_pending").await.unwrap();
+    /// assert!(matches!(status, UpdateStatus::Processed { .. }));
+    /// # });
+    /// ```
+    pub async fn wait_for_pending_update(
+        &self,
+        interval_ms: Option<u64>,
```
Mubelotix, line 100: Rust standard library offers a `Duration` struct. I think we should use it instead of `u64`. It's more idiomatic.

src/progress.rs
```diff
@@ -50,6 +53,92 @@ impl<'a> Progress<'a> {
...
+        while timeout > elapsed_time {
+            status = self.get_status().await?;
+
+            match status {
+                UpdateStatus::Failed { .. } | UpdateStatus::Processed { .. } => {
+                    return self.get_status().await;
+                },
+                UpdateStatus::Enqueued { .. } => {
+                    elapsed_time += interval;
+                    sleep(Duration::from_millis(interval));
```
Mubelotix, line 128: This sleep function is blocking. It should be async since this function is async. It will be hard without adding a new dependency... On wasm, we can use `wasm-bindgen-futures`. Do you have any idea for native?

src/progress.rs
```diff
@@ -50,6 +53,92 @@ impl<'a> Progress<'a> {
...
+                },
+            };
+        }
+
+        Err(
```
Mubelotix, line 133: It would be better to return None in the case of a timeout. It's much easier to handle. Additionally, a timeout is a normal behavior of the server and cannot be considered an error.

Mubelotix (changes requested):

src/progress.rs
sanders41: Good point. I’m not sure of a way to do this with the standard library either.

src/progress.rs
sanders41: I did this (and added the error type) because #77 requested it:

> If the method times out, it should return a TimeOut Exception.

If the team if good with returning None instead, I’ll do that.

src/progress.rs
sanders41: Will do

src/progress.rs
sanders41: I've done a bit more research on this and the only solutions I am finding are to use the sleep from either the async_std crate or tokio crate.

Mubelotix: You are right. I think it would be extremely difficult without dependency. `futures` is smaller than `async_std` and `tokio` so let's use `futures`. The dependency should be added in the `[target.'cfg(not(target_arch = "wasm32"))'.dependencies]` section of `Cargo.toml`, near `isahc`. Then you will have to use `cfg`s to create a custom async sleep function.

On native:
- Spawn a future
- Create a thread which owns an async oneshot channel sender
- Asynchronously wait for data in the receiver part of the channel, outside the thread

On wasm:
Copy and improve [this sleep function](https://github.com/Mubelotix/wasm-extension-template/blob/master/src/util.rs). `js-sys` dependency will need to be added but it's already used by web-sys so the executable size should not be increased.

src/progress.rs
Mubelotix: You are right. What do you think is best @curquiza?

sanders41: This is what I have so far for the native. It is passing the tests, but this is also my first time using the Futures crate so I have a few questions.

```rust
use std::thread::sleep;
use std::time::Duration;
use futures::channel::oneshot::{channel, Sender};
...

pub async fn wait_for_pending_update(
        &self,
        interval_ms: Option<Duration>,
        timeout_ms: Option<Duration>,
    ) -> Result<UpdateStatus, Error> {
    ...
                UpdateStatus::Enqueued { .. } => {
                    elapsed_time += interval;
                    let (sender, receiver) = channel::<()>();
                    let _thread = std::thread::spawn(move || async_sleep(sender, interval));
                    receiver.await.unwrap();
                },
    ...
}

#[cfg(not(target_arch = "wasm32"))]
pub(crate) fn async_sleep(sender: Sender<()>, interval: Duration) {
    sleep(interval);
    sender.send(()).unwrap();
}
```

1. Does this look correct, or would you expect something different?
2. Is it enough to send the sleep into the channel, or do I need to send the whole test?
2. Do you know of a way I can confirm that this isn't blocking?

src/progress.rs
sanders41: This is what I have so far for the native. It is passing the tests, but this is also my first time using the Futures crate so I have a few questions.

```rust
use std::thread::sleep;
use std::time::Duration;
use futures::channel::oneshot::{channel, Sender};
...

pub async fn wait_for_pending_update(
        &self,
        interval_ms: Option<Duration>,
        timeout_ms: Option<Duration>,
    ) -> Result<UpdateStatus, Error> {
    ...
                UpdateStatus::Enqueued { .. } => {
                    elapsed_time += interval;
                    let (sender, receiver) = channel::<()>();
                    let _thread = std::thread::spawn(move || async_sleep(sender, interval));
                    receiver.await.unwrap();
                },
    ...
}

#[cfg(not(target_arch = "wasm32"))]
pub(crate) fn async_sleep(sender: Sender<()>, interval: Duration) {
    sleep(interval);
    sender.send(()).unwrap();
}
```

1. Does this look correct, or would you expect something different?
2. Is it enough to send the sleep into the channel, or do I need to send the whole loop?
2. Do you know of a way I can confirm that this isn't blocking?

Mubelotix: Great ! It looks correct and that's what I was expecting. It suggest you move the whole workaround in the async_sleep function, including the thread so that you get a nicer interface usable like this :
```rust
UpdateStatus::Enqueued { .. } => {
    elapsed_time += interval;
    async_sleep(sender, interval).await;
},
```
You will also need to move the specific `use` keywords in the function since wasm will have no access to the `future` dependency.

You could do some logging in multiple futures `join!`ed together to confirm that it is not blocking but it should be visible in the code. The only blocking function called is wrapped in the new thread so the main thread cannot be blocked in any way.

sanders41: After making this update so it can be called with `async_sleep(interval).await;` it is now:

```rust
#[cfg(not(target_arch = "wasm32"))]
pub(crate) async fn async_sleep(interval: Duration) {
    let (sender, receiver) = futures::channel::oneshot::channel::<()>();
    let _thread = std::thread::spawn(move || std::thread::sleep(interval));
    sender.send(()).unwrap();
    let _ = async { receiver.await };
}
```

This function and the wasm one are currently in the progress.rs file, but even though they are currently only used here it seems possible they could be used in other places in the future. With that should they be moved to a different location? Maybe a sleep.rs file, utils.rs file, or something like this?

Mubelotix: That's a lot cleaner thanks! You are right about moving this function; that could be useful in the future. However, I doubt there will be other usages in the crate. Maybe you could even bury the function inside `wait_for_pending_update`. Anyway, we will remember the existence of this code and will use it if needed.
Feel free to ask for help while building the Wasm version.

There is a little problem with this code. We have to send data through the channel **from** the new thread, after the sleep.

sanders41: For the wasm version I currently have:

```rust
#[cfg(target_arch = "wasm32")]
pub(crate) async fn async_sleep(interval: Duration) {
    use wasm_bindgen_futures::JsFuture;
    
    JsFuture::from(js_sys::Promise::new(&mut |yes, _| {
        web_sys::window()
            .unwrap()
            .set_timeout_with_callback_and_timeout_and_arguments_0(
                &yes,
                interval.as_millis() as i32,
            )
            .unwrap();
    })).await.unwrap();
}
```

When I try to run the tests to see what needs updating with `cargo test --target=wasm32-unknown-unknown -- --test-threads=1` I get an error `cannot execute binary file` (this error happens even without the new function present). I have found other people getting this error, but no one with a solution. Any ideas?

Mubelotix: There is no way to test the wasm executable out of the box. You have to call the code from a complete wasm project to test things. I have made [an example](https://github.com/meilisearch/meilisearch-rust/tree/main/examples/web_app) of using MeiliSearch in wasm. It should be a great resource to start since everything is ready to be compiled and served.

sanders41:
I went ahead and pushed the non-blocking sleep update. If the decision is made to return None instead of a time out error I can still make that change also.

src/progress.rs
curquiza: If it's more rust idiomatic, let's do the `None` :)

curquiza: (sorry for the delay)

sanders41: OK, what i ended up doing is wrapping the whole result in an Option. My thinking was you would want to be able to differentiate between the wait taking longer than specified and an actual server error.

If you all disagree with that and would rather always return None let me know and I'll change it.

src/progress.rs
```diff
@@ -50,6 +52,117 @@ impl<'a> Progress<'a> {
         )
         .await
     }
+
+    /// Wait until MeiliSearch processes an update, and get its status.
+    /// 
+    /// interval_ms = The frequency at which the server should be polled. Default = 50ms
+    /// timeout_ms = The maximum time to wait for processing to complete. Default = 5000ms
```
Mubelotix, line 59:
```suggestion
    /// `interval` = The frequency at which the server should be polled. Default = 50ms
    /// `timeout` = The maximum time to wait for processing to complete. Default = 5000ms
```
Let's remove the `ms` suffix as we switched to `Duration` without unit.

src/progress.rs
```diff
@@ -50,6 +52,117 @@ impl<'a> Progress<'a> {
...
+    /// let status = progress.wait_for_pending_update(None, None).await.unwrap();
+    /// 
+    /// # client.delete_index("movies_wait_for_pending").await.unwrap();
+    /// assert!(matches!(status.unwrap(), UpdateStatus::Processed { .. }));
+    /// # });
+    /// ```
+    pub async fn wait_for_pending_update(
+        &self,
+        interval_ms: Option<Duration>,
+        timeout_ms: Option<Duration>,
```
Mubelotix, line 102:
```suggestion
        interval: Option<Duration>,
        timeout: Option<Duration>,
```

src/progress.rs
```diff
@@ -50,6 +52,117 @@ impl<'a> Progress<'a> {
...
+
+        match interval_ms {
+            Some(v) => interval = v,
+            None => interval = Duration::from_millis(50),
+        }
+
+        match timeout_ms {
+            Some(v) => timeout = v,
+            None => timeout = Duration::from_millis(5000),
+        }
```
Mubelotix, line 115:
```suggestion
        let interval = interval.unwrap_or(Duration::from_millis(50));
        let timeout = timeout.unwrap_or(Duration::from_millis(5000));
```

src/progress.rs
```diff
@@ -50,6 +52,117 @@ impl<'a> Progress<'a> {
...
+#[cfg(target_arch = "wasm32")]
+pub(crate) async fn async_sleep(interval: Duration) {
+    use wasm_bindgen_futures::JsFuture;
+
+    JsFuture::from(js_sys::Promise::new(&mut |yes, _| {
+        web_sys::window()
+            .unwrap()
+            .set_timeout_with_callback_and_timeout_and_arguments_0(
+                &yes,
+                interval.as_millis() as i32,
```
Mubelotix, line 162:
```suggestion
                interval.as_millis().try_into().unwrap(),
```
Actually casting `u64` to `i32` can be dangerous so we should panic in case of overflow.

Mubelotix (changes requested):
That's almost perfect ! I still have a few suggestions before merging.

src/progress.rs
sanders41: @Mubelotix this update triggers a clippy warning ```use of `unwrap_or` followed by a function call```. I can resolve the warning by instead using

```rust
let interval = interval.unwrap_or_else(|| Duration::from_millis(50));
let timeout = timeout.unwrap_or_else(|| Duration::from_millis(5000));
```

Does this look good to you? All other suggestions are added and ready to go.

Mubelotix: Yes that's great

Mubelotix (approved):
Awesome!

curquiza (approved):
Thanks to both of you! 🚀 

bors merge

PR #112, opened 2021-02-20
Improve settings

### LINKED ISSUE #38: Settings can only store owned data
The `Settings` struct is using `Vec`s and `String`s, which are owned structures. 
It should also be able to use borrowed values.
We should replace these owned types by generics (but since there is many fields it would make a lot of generics 😬).

Mubelotix:
Allow users to use a variety of data types when constructing `Settings` with the builder syntax.
Fix #38.


curquiza (approved):
Awesome!!! Thanks a lot!

curquiza:
bors merge

curquiza:
@Mubelotix can you confirm this PR is breaking? (for the semver)

Mubelotix:
Yes it's breaking

PR #119, opened 2021-03-24
Update health method

curquiza:
MeiliSearch will return `200` with `{ "status": "available" }` (in the v0.20.0) that is releasing next Monday 🙂

I tried to update the code accordingly @Mubelotix, tell me if I did something wrong. The `Empty` error seemed to be there only as a workaround for the `204`.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -250,11 +250,9 @@ pub struct ClientStats {
 ///
 /// ```
 /// # use meilisearch_sdk::{client::*, indexes::*, errors::Error};
-/// # futures::executor::block_on(async move {
 /// Health {
 ///    status: "available".to_string(),
 /// }
-/// # });
 /// ```
 #[derive(Deserialize)]
 pub struct Health {
--- a/src/client.rs
+++ b/src/client.rs
@@ -274,13 +274,11 @@ pub struct Keys {
 ///
 /// ```
 /// # use meilisearch_sdk::{client::*, indexes::*, errors::Error};
-/// # futures::executor::block_on(async move {
 /// Version {
 ///    commit_sha: "b46889b5f0f2f8b91438a08a358ba8f05fc09fc1".to_string(),
 ///    build_date: "2019-11-15T09:51:54.278247+00:00".to_string(),
 ///    pkg_version: "0.1.1".to_string(),
 /// }
-/// # });
 /// ```
 #[derive(Deserialize)]
 #[serde(rename_all = "camelCase")]
--- a/src/client.rs
+++ b/src/client.rs
@@ -180,30 +180,17 @@ impl<'a> Client<'a> {
     /// #
     /// # futures::executor::block_on(async move {
     /// let client = Client::new("http://localhost:7700", "masterKey");
-    ///
-    /// match client.health().await {
-    ///     Ok(()) => println!("server is operational"),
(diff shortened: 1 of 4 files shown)


src/client.rs
```diff
@@ -257,6 +244,12 @@ pub struct ClientStats {
     pub indexes: HashMap<String, IndexStats>,
 }
 
+#[derive(Deserialize)]
+#[serde(rename_all = "camelCase")]
```
Mubelotix, line 248: The unique field is already in camelCase anyway

```suggestion
```

src/client.rs
```diff
@@ -257,6 +244,12 @@ pub struct ClientStats {
...
+pub struct Health {
+    pub status: Option<String>,
```
Mubelotix, line 250: It would be great to provide some basic documentation about `Health` and the value of its field.

Mubelotix (dismissed):
That's the perfect solution! Only documentation and readability could be improved.

curquiza:
I've added documentation to `Health` structure. Telle me if it's what you expected @Mubelotix 🙂

src/client.rs
```diff
@@ -257,6 +244,18 @@ pub struct ClientStats {
     pub indexes: HashMap<String, IndexStats>,
 }
 
+/// Health of the MeiliSearch server.
+/// Example:
+/// ```text
```
Mubelotix, line 249: It's not text but Rust code. `rust` is the default so we can omit the identifier of the language. 

```suggestion
/// ```
```

src/client.rs
```diff
@@ -257,6 +244,18 @@ pub struct ClientStats {
     pub indexes: HashMap<String, IndexStats>,
 }
 
+/// Health of the MeiliSearch server.
+/// Example:
```
Mubelotix, line 248: We usually document like this

```suggestion
/// Health of the MeiliSearch server.
/// 
/// # Example
/// 
```

src/client.rs
```diff
@@ -257,6 +244,18 @@ pub struct ClientStats {
...
+/// ```text
+/// Health {
+///    status: "available".to_string(),
```
Mubelotix, line 251: You used the `status` field as if it was not optional here.
I don't know where to find documentation for the upcoming meilisearch version so I was not able to check if it is really nullable/optional. But you must either remove the `Option` at line 256 either use `Some("available".to_string())` at line 251.

Mubelotix (changes requested):

src/client.rs
curquiza: I took the only bad example in the file 😂 
https://github.com/meilisearch/meilisearch-rust/blob/c860dd0af4bd5e8a6ca412026a1a125594cbfd9a/src/client.rs#L267-L275

curquiza: I'll change it this week 🙂

src/client.rs
curquiza: It's not an `Option`, you're right! I went too fast!

curquiza:
When you have time, can you give a last check to this PR @Mubelotix? 😇 I tried to follow your requirements

Mubelotix (changes requested):
The examples are no `async` so you can remove a few lines. The rest is perfect!

PR #120, opened 2021-03-25
Type flexibility improvements

Mubelotix:
[Rust 1.51 is stable!](https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html)
That allows us to build much more powerful traits using const generics so that this library can handle a huge number of types!


curquiza (approved):
Awesome! Thank you for making this package up to date 🙂

bors merge

curquiza:
I add the label `breaking-change` since it changes the usage

PR #124, opened 2021-04-19
Standardize health method 

alallema:
**Description**
Checking that method `health()` return `{'status': 'available'}` and added `is_healthy()` method who return boolean value

**Issue related**
meilisearch/integration-guides#55

**Some points of concern**
- I let `Health` struct like it was:
```rust
#[derive(Deserialize)]
pub struct Health {
    pub status: String,
}
```
but I see we could write it this way too:
```rust
#[derive(Deserialize)]
#[serde(tag="status")]
pub enum Health {
    #[serde(rename="available")]
    Available
}
```
 That's means API should return `'available'` and this could lead to more maintenance but it seemed to be more accurate.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -203,6 +203,7 @@ impl<'a> Client<'a> {
     /// # futures::executor::block_on(async move {
     /// let client = Client::new("http://localhost:7700", "masterKey");
     /// let health = client.is_healthy().await;
+    /// assert_eq!(health, true);
     /// # });
     /// ```
     pub async fn is_healthy(&self) -> bool {
--- a/src/client.rs
+++ b/src/client.rs
@@ -198,7 +198,7 @@ impl<'a> Client<'a> {
     /// # Example
     ///
     /// ```
-    /// # use meilisearch_sdk::{client::*};
+    /// # use meilisearch_sdk::client::*;
     /// #
     /// # futures::executor::block_on(async move {
     /// let client = Client::new("http://localhost:7700", "masterKey");
--- a/src/client.rs
+++ b/src/client.rs
@@ -207,7 +207,7 @@ impl<'a> Client<'a> {
     /// # });
     /// ```
     pub async fn is_healthy(&self) -> bool {
-        if let Some(health) = self.health().await {
+        if let Ok(health) = self.health().await {
             health.status.as_str() == "available"
         } else {
             false
--- a/src/client.rs
+++ b/src/client.rs
@@ -207,7 +207,11 @@ impl<'a> Client<'a> {
     /// # });
     /// ```
     pub async fn is_healthy(&self) -> bool {
-        self.health().await.is_ok()
+        if let Some(health) = self.health().await {
+            health.status.as_str() == "available"
+        } else {
+            false
+        }
     }
 
     /// Get the private and public key.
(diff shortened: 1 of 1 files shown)


src/client.rs
```diff
@@ -193,6 +193,22 @@ impl<'a> Client<'a> {
         .await
     }
 
+    /// Get health of MeiliSearch server, return true or false.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use meilisearch_sdk::{client::*};
```
Mubelotix, line 201:
```suggestion
    /// # use meilisearch_sdk::client::*;
```

src/client.rs
```diff
@@ -193,6 +193,22 @@ impl<'a> Client<'a> {
...
+    /// #
+    /// # futures::executor::block_on(async move {
+    /// let client = Client::new("http://localhost:7700", "masterKey");
+    /// let health = client.is_healthy().await;
+    /// # });
```
Mubelotix, line 206:
```suggestion
    /// let health = client.is_healthy().await;
    /// assert_eq!(health, true);
    /// # });
```

src/client.rs
```diff
@@ -193,6 +193,22 @@ impl<'a> Client<'a> {
...
+    /// ```
+    pub async fn is_healthy(&self) -> bool {
+        self.health().await.is_ok()
+    }
```
Mubelotix, line 210:
```suggestion
    pub async fn is_healthy(&self) -> bool {
        if let Some(health) = self.health().await {
            health.status.as_str() == "available"
        } else {
            false
        }
    }
```
There is a mistake here. Calling `is_ok` on the Result will return whether the HTTP request was successful. So it would return `true` almost always, even when `status` is not `available`. I think you wanted to check the content of `status`. That can be done this way.

Mubelotix (changes requested):
The health can indeed be improved. I don't think we need a health enum yet, but this is a great idea that should be applied as soon as https://github.com/meilisearch/MeiliSearch/blob/94c0858c276743e8b2f4c8c97880070a198341bf/meilisearch-http/src/routes/health.rs#L12 stops returning "available" in all cases.

src/client.rs
alallema: @Mubelotix  This is exactly what I want to do. Thanks for your help!

alallema:
@Mubelotix Thanks again for your fixs and your help!

src/client.rs
```diff
@@ -193,6 +193,27 @@ impl<'a> Client<'a> {
...
+    /// # use meilisearch_sdk::client::*;
+    /// #
+    /// # futures::executor::block_on(async move {
+    /// let client = Client::new("http://localhost:7700", "masterKey");
+    /// let health = client.is_healthy().await;
+    /// assert_eq!(health, true);
+    /// # });
+    /// ```
+    pub async fn is_healthy(&self) -> bool {
+        if let Some(health) = self.health().await {
```
Mubelotix, line 210:
```suggestion
        if let Ok(health) = self.health().await {
```

Mubelotix (changes requested):
Whoops I made a small mistake too

Mubelotix (approved):
Perfect!

curquiza (approved):
bors merge

PR #125, opened 2021-04-21
Check code samples automatically

Mubelotix:
I added a small build script that will generate a test file from code samples.
The file will be located in `tests/generated_from_code_samples.rs` and contains 78 additional tests.
This is a great way to track breaking changes in the code samples.
I also made a few fixes to code samples as there was breaking changes in the past that were not updated.

<details><summary>Generated test preview</summary>
<p>

```rust

#[async_test]
async fn code_sample_delete_one_document_1() {
    // Setup
    let client = Client::new("http://localhost:7700", "masterKey");
    let _ = client.delete_index("delete_one_document_1").await;
    let movies = setup_test_index(&client, "delete_one_document_1").await;
    
    // Code sample
    let progress: Progress = movies.delete_document(25684).await.unwrap();

    // Cleanup
    let _ = client.delete_index("delete_one_document_1").await;
}

#[async_test]
async fn code_sample_delete_documents_1() {
    // Setup
    let client = Client::new("http://localhost:7700", "masterKey");
    let _ = client.delete_index("delete_documents_1").await;
    let movies = setup_test_index(&client, "delete_documents_1").await;
    
    // Code sample
    let progress: Progress = movies.delete_documents(&[23488, 153738, 437035, 363869]).await.unwrap();

    // Cleanup
    let _ = client.delete_index("delete_documents_1").await;
}

// and 76 other tests like these 2
```

</p>
</details>

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -21,11 +21,12 @@ get_documents_1: |-
 add_or_replace_documents_1: |-
   let progress: Progress = movies.add_or_replace(&[
     Movie {
-      id: 287947,
+      id: "287947".to_string(),
       title: "Shazam".to_string(),
       poster: "https://image.tmdb.org/t/p/w1280/xnopI5Xtky18MPhK40cZAGAOVeV.jpg".to_string(),
       overview: "A boy is given the ability to become an adult superhero in times of need with a single magic word.".to_string(),
-      release_date: "2019-03-23".to_string(),
+      release_date: 1553323329, // 2019-03-23
+      genres: vec![],
     }
   ], None).await.unwrap();
 add_or_update_documents_1: |-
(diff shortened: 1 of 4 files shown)


Mubelotix:
That's a lot of tests actually.
Maybe we can avoid running code samples test. A compilation check should be enough for them in most cases but I fear it would not be perfect.

Mubelotix:
What do you think about this @curquiza? Should we run code samples or just check that they compile? Running adds a few minutes of computation on the CI.

curquiza:
It's as you wish :)
This is too much work for the Meili team to integrate into all SDKs, it means what you are doing is a huge bonus in this one 😁 So, do what you prefer since the Meili team haven't made any decision about this yet 🙂 Adding only the compilation is already a good step from my POV :)

PR #133, opened 2021-05-10
Create get_update fn to get one update status

icyJoseph:
Hi,

Having fun using this SDK, but I reckon a function to get the status of one single update is missing.

When I read through https://docs.meilisearch.com/reference/api/updates.html#get-an-update-status, I see for example:

```javascript
 client.index('movies').getUpdateStatus(1)
```

Where the argument of `getUpdateStatus`, I guess, is the update id based on what the cURL version does:

```bash
curl \
  -X GET 'http://localhost:7700/indexes/movies/updates/1'
```

Which mirrors the MeiliSearch `/indexes/:index_uid/updates/:updateId` endpoint.

The rather unlucky thing is that to do the same with the Rust SDK it is not sufficient, not possible, to use the update id.

```rust
 let status: Status = progress.get_status().await.unwrap();
```

The above requires access to the `progress` instance, which is not possible in my case. Regardless of any specific use case, _**it is my opinion**_ that for completeness sake, the `get_update` function, defined in this PR, is needed to complete the SDK. Of course this can be challenged, or perhaps there's a reason to why such function does not exist.

Thanks for the SDK! Great Job!

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -59,16 +59,12 @@ search_1: |-
     .execute()
     .await
     .unwrap();
-get_update_md: |-
-  You can get the status of a `Progress` instance:
-  ```rust 
+get_update_1: |-
+  // You can get the status of a `Progress` object:
   let status: Status = progress.get_status().await.unwrap();
-  ```
 
-  Or you can use index to get an update status using its `update_id`:
-  ```rust
+  // Or you can use index to get an update status using its `update_id`:
   let status: Status = index.get_update(1).await.unwrap();
-  ```
 get_all_updates_1: |-
   let status: Vec<ProgressStatus> = index.get_all_updates().await.unwrap();
 get_keys_1: |-
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -59,8 +59,16 @@ search_1: |-
     .execute()
     .await
     .unwrap();
-get_update_1: |-
-  let status: Status = index.get_status(1).await.unwrap();
+get_update_md: |-
+  You can get the status of a `Progress` instance:
+  ```rust 
+  let status: Status = progress.get_status().await.unwrap();
+  ```
+
+  Or you can use index to get an update status using its `update_id`:
+  ```rust
+  let status: Status = index.get_update(1).await.unwrap();
+  ```
 get_all_updates_1: |-
   let status: Vec<ProgressStatus> = index.get_all_updates().await.unwrap();
 get_keys_1: |-
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -60,7 +60,7 @@ search_1: |-
     .await
     .unwrap();
 get_update_1: |-
-  let status: Status = progress.get_status().await.unwrap();
+  let status: Status = index.get_status(1).await.unwrap();
 get_all_updates_1: |-
   let status: Vec<ProgressStatus> = index.get_all_updates().await.unwrap();
 get_keys_1: |-
--- a/src/indexes.rs
+++ b/src/indexes.rs
@@ -619,6 +619,66 @@ impl<'a> Index<'a> {
         self.update(primary_key).await
     }
 
+    /// Get the status of a given updates in a given index.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use serde::{Serialize, Deserialize};
+    /// # use std::thread::sleep;
+    /// # use std::time::Duration;
+    /// # use meilisearch_sdk::{client::*, document, indexes::*, progress::UpdateStatus};
+    /// #
+    /// # #[derive(Debug, Serialize, Deserialize, PartialEq)]
+    /// # struct Document {
+    /// #    id: usize,
+    /// #    value: String,
+    /// #    kind: String,
+    /// # }
+    /// #
+    /// # impl document::Document for Document {
+    /// #    type UIDType = usize;
+    /// #
+    /// #    fn get_uid(&self) -> &Self::UIDType {
+    /// #        &self.id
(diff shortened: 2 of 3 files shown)


src/progress.rs
```diff
@@ -27,6 +27,24 @@ pub struct Progress<'a> {
 }
 
 impl<'a> Progress<'a> {
+
+    ///
```
Mubelotix, line 31:
```suggestion
```

src/indexes.rs
```diff
@@ -726,4 +800,34 @@ mod tests {
...
+        let progress = index
+        .add_or_replace(
+            &[Movie {
+                name: "test".to_string(),
+                description: "no desc".to_string(),
+            }],
+            Some("name"),
+        )
+        .await
+        .unwrap();
```
Mubelotix, line 820:
```suggestion
        let progress = index.delete_all_documents().await.unwrap();
```
In this test, we just need to get a `Progress`. `Client.delete_all_documents()` is the most compact way to do this. We can also remove the `Movie` struct.

src/indexes.rs
```diff
@@ -712,9 +772,23 @@ pub struct IndexStats {
...
+        description: String,
+    }
+
+    impl Document for Movie {
+      type UIDType = String;
+      fn get_uid(&self) -> &Self::UIDType {
+          &self.name
+        }
+    }
+
```
Mubelotix, line 791:
```suggestion
```
See next comment

src/indexes.rs
```diff
@@ -712,9 +772,23 @@ pub struct IndexStats {
 
 #[cfg(test)]
 mod tests {
-    use crate::{client::*};
+    use crate::{client::*, document::Document, progress::UpdateStatus};
+    use serde::{Deserialize, Serialize};
```
Mubelotix, line 776:
```suggestion
    use crate::{client::*, progress::UpdateStatus};
```
See next comment

src/indexes.rs
```diff
@@ -619,6 +619,66 @@ impl<'a> Index<'a> {
...
+    /// let client = Client::new("http://localhost:7700", "masterKey");
+    /// let movies = client.get_or_create("movies_get_one_update").await.unwrap();
+    ///
+    /// let progress = movies.add_documents(&[
+    ///     Document { id: 0, kind: "title".into(), value: "The Social Network".to_string() },
+    ///     Document { id: 1, kind: "title".into(), value: "Harry Potter and the Sorcerer's Stone".to_string() },
+    /// ], None).await.unwrap();
+    ///
+    /// let update_id = progress.get_update_id();
+    /// let status = movies.get_update(update_id).await.unwrap();
```
Mubelotix, line 657: I fear that some users will misunderstand this example. They might think that this is the only way to get a `Progress` and they might use this instead of `progress.get_status()`. Could you make this clearer? By telling that they should use `progress.get_status()` in most cases and explaining when this is intended to be used.

Mubelotix (changes requested):

src/indexes.rs
icyJoseph: I did a couple of changes to try and tackle this.

First in this very function description, I added a sentence that describes the normal course of action, that is, getting the status directly from the progress object.

Second, in the code example, I obtain the status in both ways, from the progress object, and by calling `get_update` on the index with the `update_id`.

I also modified (and fixed after my last commit) the `.code-samples.meilisearch.yml` file, by expanding the `get_update_md` section (formerly `get_update_1`), with code for both use cases.

I am not so good with copy or technical writing, so I was forced to introduce the wording **"progress object"** which might not be ideal.

What do you think? Does it look good now?

Mubelotix: Great! I like this documentation.

> I also modified (and fixed after my last commit) the .code-samples.meilisearch.yml file, by expanding the get_update_md section (formerly get_update_1), with code for both use cases.

That would be perfect if it worked but I am not sure that the documentation scraper is powerful enough. Will this work @curquiza ? Is the format of the documentation automatically detected from the field name suffix or is it hardcoded ? 

> I am not so good with copy or technical writing, so I was forced to introduce the wording "progress object" which might not be ideal.

This is indeed not ideal. I would rather say **"`Progress` struct"**. But the second occurrence of "object" is fine in my opinion.

.code-samples.meilisearch.yaml
```diff
@@ -59,8 +59,16 @@ search_1: |-
...
+get_update_md: |-
+  You can get the status of a `Progress` instance:
+  ```rust 
   let status: Status = progress.get_status().await.unwrap();
+  ```
+
+  Or you can use index to get an update status using its `update_id`:
+  ```rust
+  let status: Status = index.get_update(1).await.unwrap();
+  ```
```
curquiza, line 71:
```suggestion
get_update_1: |-
  // You can get the status of a `Progress` instance:
  let status: Status = progress.get_status().await.unwrap();

  // Or you can use index to get an update status using its `update_id`:
  let status: Status = index.get_update(1).await.unwrap();
```

src/indexes.rs
curquiza: Hello! It will indeed not work in the docs, but you can change it like this: https://github.com/meilisearch/meilisearch-rust/pull/133/files#r638115933

curquiza:
bors try

icyJoseph:
What happened to the `wasm-build`? is that done through another action?

curquiza:
> What happened to the wasm-build? is that done through another action?

I removed this test from the settings, my bad 🙂 #138 is not merged yet

curquiza (approved):
I approve, but I wait for the @Mubelotix approval before any merge 🙂
Thanks again for this PR @icyJoseph

Mubelotix (approved):
I approve too!

curquiza:
Awesome!!
Thanks to both of you!

bors merge

PR #142, opened 2021-06-06
Adding delete_index_if_exists method

sanders41:
Relates to https://github.com/meilisearch/integration-guides/issues/107

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -129,15 +129,15 @@ impl<'a> Client<'a> {
     /// To delete an index if it exists from the [`Index`] object, use the [Index::delete_if_exists] method.
     pub async fn delete_index_if_exists(&self, uid: &str) -> Result<bool, Error> {
         match self.delete_index(uid).await {
-            Ok (_) => return Ok(true),
+            Ok (_) => Ok(true),
             Err (Error::MeiliSearchError {
                 message: _,
                 error_code: ErrorCode::IndexNotFound,
                 error_type: _,
                 error_link: _,
-            }) => return Ok(false),
-            Err(error) => return Err(error),
-        };
+            }) => Ok(false),
+            Err(error) => Err(error),
+        }
     }
 
     /// Delete an index from its UID.
--- a/src/client.rs
+++ b/src/client.rs
@@ -126,22 +126,17 @@ impl<'a> Client<'a> {
     }
 
     /// Delete an index from its UID if it exists.
-    /// To delete an index if it exists from the [index object](../indexes/struct.Index.html), use
-    /// [the delete_if_exists method](../indexes/struct.Index.html#method.delete_if_exists).
+    /// To delete an index if it exists from the [`Index`] object, use the [Index::delete_if_exists] method.
     pub async fn delete_index_if_exists(&self, uid: &str) -> Result<bool, Error> {
         match self.delete_index(uid).await {
             Ok (_) => return Ok(true),
-            Err (error) => {
-                match error {
-                    Error::MeiliSearchError {
-                        message: _,
-                        error_code: ErrorCode::IndexNotFound,
-                        error_type: _,
-                        error_link: _,
-                    } => return Ok(false),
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -125,6 +125,26 @@ impl<'a> Client<'a> {
         .into_index(self))
     }
 
+    /// Delete an index from its UID if it exists.
+    /// To delete an index if it exists from the [index object](../indexes/struct.Index.html), use
+    /// [the delete_if_exists method](../indexes/struct.Index.html#method.delete_if_exists).
```
Mubelotix, line 130: It's better to [link to items by name](https://doc.rust-lang.org/rustdoc/linking-to-items-by-name.html).

```suggestion
    /// To delete an index if it exists from the [`Index`] object, use the [Index::delete_if_exists] method.
```

src/client.rs
```diff
@@ -125,6 +125,26 @@ impl<'a> Client<'a> {
...
+                match error {
+                    Error::MeiliSearchError {
+                        message: _,
+                        error_code: ErrorCode::IndexNotFound,
+                        error_type: _,
+                        error_link: _,
+                    } => return Ok(false),
+                    _ => return Err(error),
+                };
+            },
```
Mubelotix, line 144: Is clippy ok with this? It seems that `if let` should be used instead of `match` but it would be even better to match the inner pattern directly in the first match. I did not test the suggested code.
```suggestion
            Err (Error::MeiliSearchError {
                message: _,
                error_code: ErrorCode::IndexNotFound,
                error_type: _,
                error_link: _,
            }) => Ok(false),
            Err(error) => Err(error),
```

src/indexes.rs
```diff
@@ -82,6 +82,46 @@ impl<'a> Index<'a> {
...
+    /// ```
+    /// # use meilisearch_sdk::{client::*, indexes::*};
+    /// # futures::executor::block_on(async move {
+    /// let client = Client::new("http://localhost:7700", "masterKey");
+    /// # client.create_index("movies", None).await;
+    ///
+    /// // get the index named "movies" and delete it
+    /// let movies = client.assume_index("movies");
+    /// let mut deleted = movies.delete_if_exists().await.unwrap();
+    /// assert_eq!(deleted, true);
```
Mubelotix, line 98: It may be strange for somebody reading this example because `deleted` could be `false` if the line 93 was not here, and the line 93 is hidden in the doc. Consider hiding this line too or making line 93 visible.

src/indexes.rs
```diff
@@ -82,6 +82,46 @@ impl<'a> Index<'a> {
...
+            Err (error) => {
+                match error {
+                    Error::MeiliSearchError {
+                        message: _,
+                        error_code: ErrorCode::IndexNotFound,
+                        error_type: _,
+                        error_link: _,
+                    } => return Ok(false),
+                    _ => return Err(error),
+                };
```
Mubelotix, line 120: See another comment above

Mubelotix (changes requested):
Thanks for this PR! It will be much more idiomatic and makes it possible to handle errors better.

Previously, we used 
```rust
let _ = index.delete();
```
a lot but it ignores all errors which wasn't great.

src/client.rs
sanders41: Clippy gave no errors here, but I also don't mind changing it. I will try out your updated.

sanders41: The compiler complained about no return statements, but other than that this change worked. I added the returns and then made this same change in both places.

Mubelotix: Yeah that's because there is a semicolon after the match. Try removing the semicolon and you will be able to remove the returns.

sanders41: Yep, I should have noticed that...it's been a long week 😄 . Returns are removed now.

Mubelotix (approved):
Thank you @sanders41!

curquiza:
Thank you so much to both of you!!

bors merge

PR #150, opened 2021-07-17
Some MeiliSearch v0.21.0 related changes

juzi5201314:
https://github.com/meilisearch/integration-guides/issues/117
change: 
- Ranaming of buildDate into commitDate
- New update status value
- attributesForFaceting is renamed into filterableAttributes
- Rename fieldsDistribution into fieldDistribution
- filters is renamed into filter
- facetFilters is removed

### DIFF
--- a/src/progress.rs
+++ b/src/progress.rs
@@ -71,12 +71,12 @@ impl<'a> Progress {
     }
 
     /// Wait until MeiliSearch processes an update, and get its status.
-    /// 
+    ///
     /// `interval` = The frequency at which the server should be polled. Default = 50ms
     /// `timeout` = The maximum time to wait for processing to complete. Default = 5000ms
-    /// 
+    ///
     /// If the waited time exceeds `timeout` then `None` will be returned.
-    /// 
+    ///
     /// # Example
     ///
     /// ```
@@ -89,7 +89,7 @@ impl<'a> Progress {
     /// #    value: String,
     /// #    kind: String,
     /// # }
-    /// # 
+    /// #
     /// # impl document::Document for Document {
     /// #    type UIDType = usize;
     /// #
@@ -101,14 +101,14 @@ impl<'a> Progress {
     /// # futures::executor::block_on(async move {
     /// let client = Client::new("http://localhost:7700", "masterKey");
     /// let movies = client.create_index("movies_wait_for_pending", None).await.unwrap();
-    /// 
+    ///
     /// let progress = movies.add_documents(&[
     ///     Document { id: 0, kind: "title".into(), value: "The Social Network".to_string() },
     ///     Document { id: 1, kind: "title".into(), value: "Harry Potter and the Sorcerer's Stone".to_string() },
     /// ], None).await.unwrap();
-    /// 
+    ///
     /// let status = progress.wait_for_pending_update(None, None).await.unwrap();
-    /// 
+    ///
     /// # client.delete_index("movies_wait_for_pending").await.unwrap();
     /// assert!(matches!(status.unwrap(), UpdateStatus::Processed { .. }));
     /// # });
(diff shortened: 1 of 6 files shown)


src/search.rs
```diff
@@ -144,14 +144,10 @@ pub struct Query<'a> {
...
-    /// Filters applied to documents.
+    /// Filter applied to documents.
     /// Read the [dedicated guide](https://docs.meilisearch.com/reference/features/filtering.html) to learn the syntax.
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub filters: Option<&'a str>,
-    /// Facet names and values to filter on.
-    /// Read [this page](https://docs.meilisearch.com/reference/features/search_parameters.html#facet-filters) for a complete explanation.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub facet_filters: Option<&'a [&'a [&'a str]]>,
+    pub filter: Option<&'a str>,
```
bidoubiwa, line 150: The type of fillter is the same as `facet_filters` but can also accept a `str` as value. So it's a mix of both `filter` and `facet_filters`. 
The following syntaxes are possible: 

nested array:
```json
["something > 1", "genre=comedy", ["genre=horror", "title=batman"]]
```
only string: 
```json
"something > 1 AND genre=comedy AND (genre=horror OR title=batman)"
```

of course both are also mixable
```json
["something > 1 AND genre=comedy", "genre=horror OR title=batman"]
```

src/search.rs
```diff
@@ -222,15 +217,8 @@ impl<'a> Query<'a> {
...
-        self.filters = Some(filters);
-        self
-    }
-    pub fn with_facet_filters<'b>(
-        &'b mut self,
-        facet_filters: &'a [&'a [&'a str]],
-    ) -> &'b mut Query<'a> {
-        self.facet_filters = Some(facet_filters);
+    pub fn with_filter<'b>(&'b mut self, filter: &'a str) -> &'b mut Query<'a> {
+        self.filter = Some(filter);
```
bidoubiwa, line 221: Should be updated based on type update.

src/search.rs
```diff
@@ -364,53 +352,27 @@ mod tests {
...
-
-        let mut query = Query::new(&index);
-        query.with_facet_filters(&[&["kind:text", "kind:title"]]);
-        let results: SearchResults<Document> = index.execute_query(&query).await.unwrap();
-        assert_eq!(results.hits.len(), 10);
-
-        client
-            .delete_index("test_query_facet_filters")
-            .await
-            .unwrap();
```
bidoubiwa, line 413: Could you do these tests on `filter` ?

src/search.rs
```diff
@@ -434,7 +396,7 @@ mod tests {
 
         let mut query = Query::new(&index);
         query.with_facets_distribution(Selectors::Some(&["kind"]));
-        query.with_facet_filters(&[&["kind:text"]]);
+        //query.with_facet_filters(&[&["kind:text"]]);
```
bidoubiwa, line 399: should be uncommented

src/progress.rs
```diff
@@ -200,18 +200,29 @@ pub struct SettingsUpdate {
...
+        #[serde(skip_serializing_if = "Option::is_none")]
+        number: Option<usize>,
+    },
+    DocumentsDeletion {
+        #[serde(skip_serializing_if = "Option::is_none")]
+        number: Option<usize>,
+    },
+    Settings {
+        settings: SettingsUpdate,
+    },
```
bidoubiwa, line 225: Thanks a lot for raising this change, was this documented somewhere?

I have made an issue to know if this is the expected behavior or not: see meilisearch/MeiliSearch/issues/1517

bidoubiwa (changes requested):
Hello, thanks so much for this pr 🔥 Small changes that were not included in this PR: 

- `settings.rs` lines 298, 495, 698 `[attributes for faceting]` in the comments should be changed with `filterable attributes`
- `settings.rs` lines 310, 513, 715 should have their API route updated `&format!("{}/indexes/{}/settings/attributes-for-faceting", self.host, self.uid),` to `/filterable-attributes`
- `.code-samples.meilisearch.yaml` 511, 535 `with_facet_filters` should be removed

src/search.rs
bidoubiwa: After some thought we may also just only accept the string format. What do you think?

curquiza:
Since we don't have answer of the contributor, I merge and I will apply the changes myself :)

PR #170, opened 2021-09-01
Make MatchRange fields public

adaszko:
[...] so that it's possible to interpret them beyond what highlighting the SDK
supports.

Hi 👋  I found this to be getting in the way of displaying search results other than the `<em>` highlighting that the SDK supports.


curquiza (approved):
thanks a lot! sorry for the delay

bors merge

PR #205, opened 2021-11-08
`formatted_result` is now `Option<Map<String, Value>>` instead of `Option<T>`

curquiza:
Now, MeiliSearch returns an object in `_formatted` that is not the exact typed document since the number will be converted into string.


curquiza:
I did this for the moment, but I opened an issue related to my implementation here: https://github.com/meilisearch/meilisearch-rust/issues/206

PR #227, opened 2022-02-09
Api key

irevoire:
Implements the API key:
- The `get_key`, `delete_key`, `update_key` and `create_key` methods have been implemented with unit and doc test
- The `authenticate` error code has been renamed into `auth`
- 6 new errors have been added: `MissingParameter`, `InvalidApiKeyDescription`, `InvalidApiKeyActions`, `InvalidApiKeyIndexes`, `InvalidApiKeyExpiresAt`, `ApiKeyNotFound`
- The `Key` type has been moved to a different module
- A `KeyBuilder` type has been added to eases the creation of keys
- An `Action` enum has been created to represents all the [possible action](https://docs.meilisearch.com/reference/api/keys.html#create-a-key)

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,6 +1,7 @@
 use crate::{
     errors::*,
     indexes::*,
+    key::{Key, KeyBuilder},
     request::*,
     tasks::{async_sleep, Task},
     Rc,
@@ -274,17 +275,20 @@ impl Client {
         }
     }
 
-    /// Get the API keys from Meilisearch.
+    /// Get the API [Key]s from Meilisearch.
+    /// See the [meilisearch documentation](https://docs.meilisearch.com/reference/api/keys.html#get-all-keys).
+    ///
+    /// See also [Client::create_key] and [Client::get_key].
     ///
     /// # Example
     ///
     /// ```
-    /// # use meilisearch_sdk::{client::*, errors::Error};
+    /// # use meilisearch_sdk::{client::*, errors::Error, key::KeyBuilder};
     /// #
     /// # futures::executor::block_on(async move {
     /// let client = Client::new("http://localhost:7700", "masterKey");
     /// let keys = client.get_keys().await.unwrap();
-    /// assert_eq!(keys.len(), 2);
+    /// assert!(keys.len() >= 2);
     /// # });
     /// ```
     pub async fn get_keys(&self) -> Result<Vec<Key>, Error> {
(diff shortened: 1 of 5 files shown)


src/client.rs
```diff
@@ -274,17 +275,20 @@ impl Client {
...
     ///
     /// ```
-    /// # use meilisearch_sdk::{client::*, errors::Error};
+    /// # use meilisearch_sdk::{client::*, errors::Error, key::KeyBuilder};
     /// #
     /// # futures::executor::block_on(async move {
     /// let client = Client::new("http://localhost:7700", "masterKey");
     /// let keys = client.get_keys().await.unwrap();
-    /// assert_eq!(keys.len(), 2);
+    /// assert!(keys.len() > 2);
```
sanders41, line 291: I think this test is tightly coupled with the order the tests are run? By default 2 keys are created so it would only be > 2 if a key is created and still present before this test runs correct? Should it be `assert!(keys.len() >= 2);`? There is also the potential it is < 2 if a key was deleted and not re-added before this test runs.

irevoire: Oh yes good catch, I wanted to write `>= 2`!
> There is also the potential it is < 2 if a key was deleted and not re-added before this test runs.

I think for now we’re going to consider that we never delete the two default keys

src/key.rs
```diff
@@ -0,0 +1,236 @@
...
+    fn as_ref(&self) -> &str {
+        &self.key
+    }
+}
+
+impl AsRef<Key> for Key {
+    fn as_ref(&self) -> &Key {
+        self
+    }
+}
```
Kerollmops, line 36: Why do you o this, here? Is it useful somewhere?

src/key.rs
```diff
@@ -0,0 +1,236 @@
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use meilisearch_sdk::key::{KeyBuilder, Action};
+    /// let mut builder = KeyBuilder::new("My little lovely test key");
+    /// builder.with_actions(vec![Action::Search, Action::DocumentsAdd]);
+    /// ```
+    pub fn with_actions(&mut self, actions: impl IntoIterator<Item = Action>) -> &mut Self {
+        self.actions = actions.into_iter().collect();
```
Kerollmops, line 94:
```suggestion
        self.actions.extend(actions);
```

src/key.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub fn new(description: impl AsRef<str>) -> KeyBuilder {
+        Self {
+            actions: Vec::new(),
+            description: description.as_ref().to_string(),
+            expires_at: None,
+            indexes: Vec::new(),
+        }
+    }
+
+    /// Set all the actions the [Key] will be able to execute.
```
Kerollmops, line 84:
```suggestion
    /// Declare a set of actions the [Key] will be able to execute.
```

src/key.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub async fn create(&self, client: &Client) -> Result<Key, Error> {
+        client.create_key(self).await
+    }
+}
+
+impl AsRef<KeyBuilder> for KeyBuilder {
+    fn as_ref(&self) -> &KeyBuilder {
+        self
+    }
+}
```
Kerollmops, line 184: Same here, do you need to implement that?

src/request.rs
```diff
@@ -108,6 +118,11 @@ pub(crate) async fn request<
         Method::Delete => {
             request.method("DELETE");
         }
+        Method::Patch(body) => {
+            request.method("PATCH");
+            headers.append("Content-Type", "application/json").unwrap();
```
Kerollmops, line 123:
```suggestion
            headers.append(header::CONTENT_TYPE, "application/json").unwrap();
```

Kerollmops (changes requested):
Looks good to me, thank you Tamo!

src/key.rs
irevoire: Oh that’s smooth

Kerollmops (approved):
Looks good to me, merging!
bors merge

irevoire:
bors merge

bidoubiwa:
bors merge

irevoire:
bad bors

brunoocasali:
bors merge

PR #232, opened 2022-02-10
refactorise the errors

irevoire:
- Merge the two meilisearch error types
- Use Deserialize and Serialize to generate the string associated with every error instead of writing it by hand


brunoocasali:
Was this closed unintentionally @irevoire?

Issue #226, opened 2022-02-08
Use chrono


irevoire:
Use chrono to deserialise everything related to date and timestamp

irevoire:
How is it not maintained? I can’t find anything about it?
And since [MeiliSearch already use it](https://github.com/meilisearch/meilisearch/blob/main/meilisearch-lib/Cargo.toml#L15), if there is any problem with chrono it’s going to crash MeiliSearch before it crashes the sdk no?

Mubelotix:
Look at their inactive github repo. Nobody answers issues and nobody fixes them. Dependencies are critically outdated. A vulnerability was found in chrono a few months ago and extreme fear began to spread in the rust community. I'm not even sure it's fixed now. I really hope this problem is solved soon but as of today, I don't think one should use chrono.

irevoire:
Hey @Mubelotix, I was scrolling on the chrono issues, and I’ve seen most people are moving to [time](https://github.com/time-rs/time); what do you think of this one?
From what I see, it should be enough for us, and chrono was actually depending on it.

Mubelotix:
> Hey @Mubelotix, I was scrolling on the chrono issues, and I’ve seen most people are moving to [time](https://github.com/time-rs/time); what do you think of this one? From what I see, it should be enough for us, and chrono was actually depending on it.

Hi @irevoire, sorry for the late reply; notification got lost. That's exactly what I was going to suggest after reading [this thread](https://www.reddit.com/r/rust/comments/ts84n4/chrono_or_time_03/?utm_medium=android_app&utm_source=share)

irevoire:
Well now it’s done 👍

PR #237, opened 2022-02-14
Put type on timestamp and duration

irevoire:
Parse the Duration and timestamp with the `iso8601-duration` and the `time` crate

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -690,7 +690,7 @@ get_all_keys_1: |-
 create_a_key_1: |-
   let mut key_options = KeyBuilder::new("Add documents: Products API key");
   key_options.with_action(Action::DocumentsAdd)
-      .with_expires_at("2042-04-02T00:42:42Z")
+      .with_expires_at(time::macros::datetime!(2042 - 04 - 02 00:42:42 UTC))
       .with_index("products");
   let new_key = client.create_key(key_options).await.unwrap();
 update_a_key_1: |-
@@ -718,7 +718,7 @@ security_guide_create_key_1: |-
   let client = Client::new("http://localhost:7700", "masterKey");
   let mut key_options = KeyBuilder::new("Search patient records key");
   key_options.with_action(Action::Search)
-      .with_expires_at("2023-01-01T00:00:00Z")
+      .with_expires_at(time::datetime!(2023-01-01 00:00:00 UTC))
       .with_index("patient_medical_records");
   let new_key = client.create_key(key_options).await.unwrap();
 security_guide_list_keys_1: |-
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -698,7 +698,7 @@ update_a_key_1: |-
   key.description = "Manage documents: Products/Reviews API key".to_string();
   key.actions = vec![Action::DocumentsAdd, Action::DocumentsDelete];
   key.indexes = vec!["products".to_string(), "reviews".to_string()];
-  key.expires_at = Some("2042-04-02T00:42:42Z".to_string());
+  key.expires_at = Some(time::macros::datetime!(2042 - 04 - 02 00:42:42 UTC));
   let updated_key = client.update_key(&key);
 delete_a_key_1: |-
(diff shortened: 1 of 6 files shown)


irevoire:
Oops sorry I created this PR a little bit too early

irevoire:
Now it’s ready to review 👍

.code-samples.meilisearch.yaml
```diff
@@ -690,7 +690,7 @@ get_all_keys_1: |-
 create_a_key_1: |-
   let mut key_options = KeyBuilder::new("Add documents: Products API key");
   key_options.with_action(Action::DocumentsAdd)
-      .with_expires_at("2042-04-02T00:42:42Z")
+      .with_expires_at(time::datetime!(2042-04-02 00:42:42 UTC))
```
brunoocasali, line 693: You don't need the quotes (like this: `time::datetime!("2042-04-02 00:42:42 UTC")`)?

Cargo.toml
```diff
@@ -10,9 +10,11 @@ repository = "https://github.com/meilisearch/meilisearch-sdk"
 
 [dependencies]
 async-trait = "0.1.51"
-serde_json = "1.0"
+iso8601-duration = "0.1.0"
```
brunoocasali, line 13: Sorry for the newbie here 😅 

I think I get the idea, but just to be sure, the users of the rust-sdk don't need to have these time-handling libraries on their apps? 
In other words, to be able to use some function they need to have the lib, in order to have the right types to pass to the function?

brunoocasali (commented):
I have some doubts! 😅

.code-samples.meilisearch.yaml
irevoire: No, we don’t need quotes here because it’s a macro. So your date is checked at compile time directly, and `time` ensures it’s correct before running everything (and crashing at runtime).
BUT, the correct path is `time::macros::datetime` and not just `time::datetime` 😱 
Good catch!

Cargo.toml
irevoire: The `iso8601-duration` crate is never exposed to the end-user, but sadly the `time` crate will be needed if someone wants to manipulate API keys.

For the context, there was a lot of `TODO: use chrono` (which is one of the biggest crates in the rust ecosystem).
And there was this comment asking for a real « time » type instead of `String` in the SDK: https://github.com/meilisearch/integration-guides/issues/121#issuecomment-944379531

And the reason why we’re not using `chrono` is here: https://github.com/meilisearch/meilisearch-rust/issues/226
There is a bunch (2 or 3, I think) of CVE on `chrono` that are not fixed because the crate is not really maintained anymore.
But for our use case, `chrono` was just a wrapper around `time` (which is maintained and fixed all of the CVE encountered in `chrono`), so I decided to import this one instead of `chrono`.
Also, `time` is way smaller than `chrono`, and you can configure a lot of features depending on your needs.


But yes, to conclude, a user of the `meilisearch-rust` sdk can do most of the thing without any problem, but if he wants to use inspect, create or update anything related to the `OffsetDateTime`, he’ll need to import `time` in his crate.

Kerollmops (approved):
It looks good to me, thank you @irevoire!
Will let @brunoocasali merge it then :)

.code-samples.meilisearch.yaml
```diff
@@ -690,7 +690,7 @@ get_all_keys_1: |-
 create_a_key_1: |-
   let mut key_options = KeyBuilder::new("Add documents: Products API key");
   key_options.with_action(Action::DocumentsAdd)
-      .with_expires_at("2042-04-02T00:42:42Z")
+      .with_expires_at(time::macros::datetime!(2042 - 04 - 02 00:42:42 UTC))
```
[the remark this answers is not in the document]
irevoire: Ah yes sorry my code samples were wrong again 😠 
It should be fixed now

curquiza:
I put the breaking change label since it impacts the user experience (cf code sample changes) 🙂

brunoocasali:
bors merge

brunoocasali (approved):
🔥 🚒

brunoocasali:
bors merge

PR #244, opened 2022-02-22
use an OffsetDateTime on the ClientStats

irevoire:
# Pull Request

## What does this PR do?
I noticed we were still deserializing the `last_update` in the `ClientStats` structure in an `Option<String>`.
I typed the `String` into an `OffsetDateTime`.

## PR checklist

--------

I also found a "bug" in one of our doctest where we were creating an index with a bad name.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -9,6 +9,7 @@ use crate::{
 use serde::Deserialize;
 use serde_json::{json, Value};
 use std::{collections::HashMap, time::Duration};
+use time::OffsetDateTime;
 
 /// The top-level struct of the SDK, representing a client containing [indexes](../indexes/struct.Index.html).
 #[derive(Debug, Clone)]
@@ -588,7 +589,8 @@ impl Client {
 #[serde(rename_all = "camelCase")]
 pub struct ClientStats {
     pub database_size: usize,
-    pub last_update: Option<String>,
+    #[serde(with = "time::serde::rfc3339::option")]
+    pub last_update: Option<OffsetDateTime>,
     pub indexes: HashMap<String, IndexStats>,
 }
 
--- a/src/tasks.rs
+++ b/src/tasks.rs
@@ -329,7 +329,7 @@ impl Task {
     /// # futures::executor::block_on(async move {
     /// # let client = Client::new("http://localhost:7700", "masterKey");
     /// let task = client
-    ///   .create_index("is_success", None)
+    ///   .create_index("is_pending", None)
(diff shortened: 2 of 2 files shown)


src/client.rs
```diff
@@ -588,7 +589,8 @@ impl Client {
 #[serde(rename_all = "camelCase")]
 pub struct ClientStats {
     pub database_size: usize,
-    pub last_update: Option<String>,
+    #[serde(with = "time::serde::rfc3339::option")]
+    pub last_update: Option<OffsetDateTime>,
```
brunoocasali, line 593: Should I be worried because no tests were changed and, we still have the CI passing? 😅

irevoire: I totally get your concern ahah!

We actually have only [one test](https://github.com/meilisearch/meilisearch-rust/blob/main/src/client.rs#L216-L223) on this route:
```rust
use meilisearch_sdk::{client::*, indexes::*};
futures::executor::block_on(async move {
  let client = Client::new("http://localhost:7700", "masterKey");
  let stats = client.get_stats().await.unwrap();
});
```
And the `unwrap` right there ensure we were able to parse the new date format:  `client.get_stats().await.unwrap();`

------

On another note I'm wondering if we could add more tests though, but we can't test a lot of things because we don't know what the `database_size` is supposed to be, we don't really know the time of the `last_update` or what indexes are in the `Client` currently :thinking: 
```rust
pub struct ClientStats {
    pub database_size: usize,
    pub last_update: Option<String>,
    pub indexes: HashMap<String, IndexStats>,
}
```
One test we could add though would be to create an index and ensure we find it later in the stats + ensure the time of the `last_update` >= the time we sent our update but that feel like testing meilisearch more than meilisearch-rust, what do you think?

brunoocasali: I totally agree with that idea! Or maybe if it is possible you could create a mock of the API response, adding a specific value in the `database_size` and `last_update` it will make it easier to test after that. But of course, this mock need only is used in the raw response of the API, to ensure the crate will be able to parse it accordingly. 

I think this way you could have the whole TDD flow :)

irevoire: Creating a mock would be the best thing, but I don't have the time currently, sadly :unamused: 
We could open an issue, though, I think [faux](https://docs.rs/faux/latest/faux/) would be really suited for this library :+1:

brunoocasali (approved):
🔥 🔥

brunoocasali:
bors merge

PR #245, opened 2022-02-24
Create an example showing how to update the Settings

irevoire:
# Pull Request

## What does this PR do?
Provide an example that shows how to use the settings.

## PR checklist
- [ ] Does this PR fix an existing issue?

### DIFF
--- a/examples/settings.rs
+++ b/examples/settings.rs
@@ -1,0 +1,65 @@
+use meilisearch_sdk::client::Client;
+use meilisearch_sdk::indexes::Index;
+use meilisearch_sdk::settings::Settings;
+
+// we need an async runtime
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    let client: Client = Client::new("http://localhost:7700", "masterKey");
+
+    // We try to create an index called `movies` with a primary_key of `movie_id`.
+    let my_index: Index = client
+        .create_index("movies", Some("movie_id"))
+        .await
+        .expect("Could not join the remote server.")
+        // The creation of indexes is asynchronous. But for the sake of the example so we will
+        // wait until the update is entirely processed.
+        .wait_for_completion(&client, None, None)
+        .await
+        .expect("Could not join the remote server.")
+        // If the creation was successful we can generate an `Index` out of it.
+        .try_make_index(&client)
+        // This error comes from meilisearch itself.
+        .expect("An error happened with the index creation.");
+
+    // And now we can update the settings!
+    // You can read more about the available options here: https://docs.meilisearch.com/learn/configuration/settings.html#index-settings
+    let settings: Settings = Settings::new()
+        .with_searchable_attributes(["name", "title"])
+        .with_filterable_attributes(["created_at"]);
+
+    // Updating the settings is also an asynchronous operation.
+    let task = my_index
+        .set_settings(&settings)
+        .await
+        .expect("Could not join the remote server.")
+        // And here we wait for the operation to execute entirely so we can check any error happened.
+        .wait_for_completion(&client, None, None)
+        .await
+        .expect("Could not join the remote server.");
+
+    // We check if the task failed.
+    if task.is_failure() {
+        panic!(
+            "Could not update the settings. {}",
+            task.unwrap_failure().error_message
+        );
+    }
+
+    // And finally we delete the `Index`.
+    my_index
+        .delete()
+        .await
+        .expect("Could not join the remote server.")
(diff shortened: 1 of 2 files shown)


examples/settings.rs
```diff
@@ -0,0 +1,48 @@
...
+use meilisearch_sdk::indexes::Index;
+use meilisearch_sdk::settings::Settings;
+
+// we need an async runtime
+#[tokio::main(flavor = "current_thread")]
+async fn main() {
+    let client = Client::new("http://localhost:7700", "masterKey");
+
+    // We try to create an index called `movies` with a primary_key of `movie_id`.
+    let my_index: Index = client
```
brunoocasali, line 11: In line 8, you don't define the "type" but here you defined it, is this a convention or just a non-required thing?

examples/settings.rs
```diff
@@ -0,0 +1,48 @@
...
+        .await
+        .unwrap()
+        // The creation of indexes is asynchronous. But for the sake of the example so we will
+        // wait until the update is entirely processed.
+        .wait_for_completion(&client, None, None)
+        .await
+        .unwrap()
+        // If the creation was successful we can generate an `Index` out of it.
+        .try_make_index(&client)
+        .unwrap();
```
brunoocasali, line 22: I was reading some ["scientific articles"](https://stackoverflow.com/questions/36362020/what-is-unwrap-in-rust-and-what-is-it-used-for)(sarcasm) about `unwrap`. 

My question is, can we use `?` here? And can we avoid these calls or I just don't need to bother with that?

brunoocasali (commented):
Hey, just left some comments, trying to sane my doubts here!

examples/settings.rs
irevoire: Oh yes, that's useless!
I copy-pasted some code and forgot to remove it.

But now you talk about it, I think it would be good to show all the types we're using, so when someone looks at the example, he knows instantaneously which function returns what without even looking at the doc.

I'll update my PR thanks :+1:

examples/settings.rs
irevoire: Basically, we can use `?` when our function returns a `Result`.

A little rust lesson just so you know :grin: : 
```rust
let result = task.try_make_index(&client);
let index = match result {
    Ok(index) => index,
    Err(error) => return Err(error.into()), // `.into` try to convert the error to the type the current function expect
};
```

This code is equivalent to:
```rust
let index = task.try_make_index(&client)?;
```

-------------

Ok so now you know :grin: 

> My question is, can we use ? here? And can we avoid these calls or I just don't need to bother with that?

We could make our `main` return a `Result<(), meilisearch_sdk::error::Error>` (we can't return anything else than `()` in a `main`).
But in an example _I THINK_, we prefer to write the dumbest code possible to show every little step without any ambiguity.
Now that's my opinion and I can be wrong. There is no standard in rust about that.

Another thing we could do is call `expect` instead of `unwrap`. Expect allows us to write a little error message that could be more explicit for the user before unwrapping the error. Something like `Can't reach the server.`.
Like [here](https://github.com/serenity-rs/serenity/blob/current/examples/e03_struct_utilities/src/main.rs#L36-L46) for example.

examples/settings.rs
brunoocasali: Awesome! This is something I need to be used to, in some communities like in the C# they are going to the full `var` style (without explicit types). Btw there is a widely used decision regarding the theme?

> he knows instantaneously which function returns what without even looking at the doc.

I like this! 👍

examples/settings.rs
brunoocasali: Thanks a lot for the detailed explanation you're the best! Now it makes sense to me (no, I'm not an expert yet but it makes sense to me haha).
I agree with you regarding the samples, being the simplest as possible. Experienced users will not how to make it simpler :)

examples/settings.rs
irevoire: Yep usually in rust you elide all the types you can.
If the compiler is able to infer something let it do it.
Sometimes you even write « half » a type like that:
```rust
let chars: Vec<_> = "a string".chars().collect();
```
Here rust can't guess which kind of data structure you want to create with the `.collect` method so you need to say it's a `Vec`, but then you can elide the `char` because the `.chars()` method returns an iterator of `char` :grin:

brunoocasali (approved):
Good job, thanks for your contributions here @irevoire ❤️

brunoocasali:
bors merge

PR #251, opened 2022-03-20
Remove the sync feature and use Arc everywhere by default

### LINKED ISSUE #250: Make the `sync` feature enabled by default.
Currently, you can enable the `sync` feature manually however, this doesn't massively make sense to have as an opt-in thing when it's originally targeting the Tokio ecosystem where most systems will likely be using the multi-threaded runtime.

In the case that they *are* using a single-threaded runtime and are at the point of wanting to save on the atomic cost, then they can opt-out. Although the overhead of the atomic operation is completely insignificant in reality and I'm not sure if it's worth having it as a feature, to begin with considering it is so cheap.

irevoire:
# Pull Request

## What does this PR do?
Fixes #250

## PR checklist


curquiza:
bors try

curquiza:
@irevoire, not an emergency but looks like there is an issue with the wasm-build CI

irevoire:
I rebased + fixed everything 👍

irevoire:
@bidoubiwa you can merge

PR #261, opened 2022-04-14
Fixed formatting with clippy and removed Document trait

honne23:
# Pull Request

## What does this PR do?
Fixes #255 

## PR checklist
Please check if your PR fulfills the following requirements:
- [x] Does this PR fix an existing issue?
- [x] Have you read the contributing guidelines?
- [x] Have you made sure that the title is accurate and descriptive of the changes?

Thank you so much for contributing to Meilisearch!

PR #262, opened 2022-04-16
Add methods to automatically add/update documents in batches

### LINKED ISSUE #183: Add methods to automatically add/update documents in batches
We need methods to add and update the documents of MeiliSearch in batch instead of letting the users call `addDocuments` in a loop.

- [ ] Add `addDocumentsInBatches`
```
addDocumentsInBatches(documents, batchSize = 1000, primaryKey = nil)
```
`batchSize` and `primaryKey` are optional parameters

- [ ] Add `updateDocumentsInBatches`
```
updateDocumentsInBatches(documents, batchSize = 1000, primaryKey = nil)
```
`batchSize` and `primaryKey` are optional parameters

- [ ] Add tests

Example: https://github.com/meilisearch/meilisearch-python/pull/260

⚠️ This issue is generated, it means `addDocumentsInBatches` andd `updateDocumentsInBatches` might be named differently in this package. Keep the already existing way of naming in this package to stay idiomatic with the language and the repository.
Also, if you are a maintainer, feel free to add any clarification and instruction about this issue related to the specific languages of the repo.

Related to https://github.com/meilisearch/integration-guides/issues/106

Sorry if this is already partially/completely implemented, feel free to let me know about the state of this issue in the repo.

abhizer:
Fix to issue #183

# Pull Request

## What does this PR do?
Fixes #183 
https://github.com/meilisearch/meilisearch-rust/issues/183
It adds two methods, `Index::add_documents_in_batches` and `Index::update_documents_in_batches` 
Takes inspiration from the python version of the implementation. 

## PR checklist

### DIFF
--- a/src/indexes.rs
+++ b/src/indexes.rs
@@ -809,20 +809,12 @@ impl Index {
     ///
     /// ```
     /// use serde::{Serialize, Deserialize};
-    /// use meilisearch_sdk::{client::*, document::*};
+    /// use meilisearch_sdk::client::*;
     ///
     /// #[derive(Serialize, Deserialize, Debug)]
     /// struct Movie {
     ///     name: String,
     ///     description: String,
-    /// }
-    ///
-    /// impl Document for Movie {
-    ///     type UIDType = String;
-    ///
-    ///     fn get_uid(&self) -> &Self::UIDType {
-    ///         &self.name
-    ///     }
     /// }
     ///
     /// # futures::executor::block_on(async move {
@@ -855,7 +847,7 @@ impl Index {
     /// None).await.unwrap();
     /// # });
     /// ```
-    pub async fn add_documents_in_batches<T: Document>(
+    pub async fn add_documents_in_batches<T: Serialize>(
         &self,
         documents: &[T],
         batch_size: Option<usize>,
@@ -878,20 +870,12 @@ impl Index {
     ///
     /// ```
     /// use serde::{Serialize, Deserialize};
-    /// use meilisearch_sdk::{client::*, document::*};
+    /// use meilisearch_sdk::client::*;
     ///
     /// #[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
     /// struct Movie {
     ///     name: String,
     ///     description: String,
-    /// }
-    ///
-    /// impl Document for Movie {
-    ///     type UIDType = String;
-    ///
-    ///     fn get_uid(&self) -> &Self::UIDType {
-    ///         &self.name
-    ///     }
     /// }
     ///
     /// # futures::executor::block_on(async move {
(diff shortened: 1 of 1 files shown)


src/indexes.rs
```diff
@@ -875,6 +875,169 @@ impl Index {
...
+    /// # movie_index.delete().await.unwrap().wait_for_completion(&client, None,
+    /// None).await.unwrap();
+    /// # });
+    /// ```
+    pub async fn add_documents_in_batches<T: Document>(
+        &self,
+        documents: &[T],
+        batch_size: Option<usize>,
+        primary_key: Option<&str>,
+    ) -> Result<Vec<Task>, Error> {
```
irevoire, line 940:
```suggestion
    pub async fn add_documents_in_batches<T: Document>(
        &self,
        documents: impl IntoIterator<Item = T>,
        batch_size: Option<usize>,
        primary_key: Option<&str>,
    ) -> Result<Vec<Task>, Error> {
```

Hey, since you're only iterating on the documents wouldn't this work and be easier to use?

src/indexes.rs
```diff
@@ -875,6 +875,169 @@ impl Index {
...
+    /// # movie_index.delete().await.unwrap().wait_for_completion(&client, None,
+    /// None).await.unwrap();
+    /// # });
+    /// ```
+    pub async fn update_documents_in_batches<T: Document>(
+        &self,
+        documents: &[T],
+        batch_size: Option<usize>,
+        primary_key: Option<&str>,
+    ) -> Result<Vec<Task>, Error> {
```
irevoire, line 1034: Same here

irevoire (commented):
Hey, the PR looks pretty good to me but I don't really have the time to review it rn.
I will probably review it on Monday or thuesday.

Thanks for your contribution! 👍

irevoire:
Also don't bother with the clippy checks, I think we're going to merge #261 before this PR thus you will only need to rebase on main 😁

src/indexes.rs
abhizer: So something like? 
```rust
    pub async fn add_documents_in_batches<T: Document>(
        &self,
        documents: impl IntoIterator<Item = T>,
        batch_size: Option<usize>,
        primary_key: Option<&str>,
    ) -> Result<Vec<Task>, Error> {
            let documents: Vec<T> = documents.into_iter().collect();
           // unchanged
```
Just changing the type wouldn't work as methods `len` and `chunks` don't seem to be available for `Iterator`. Making `documents` trait bound would require us to collect it into a `Vec<T>`.

irevoire: Ah, sorry, I thought we had the `chunks` methods available directly on iterators. My bad!
Let's keep your code as-is then I don't want to allocate a full vector with all the documents in the back of the users :+1:

abhizer: Okay. Will rebase this once #261 is merged!

irevoire: Thanks you're the best :grin:

irevoire (approved):
Thanks for your contribution

brunoocasali:
bors merge

PR #263, opened 2022-04-18
Feature/Tenant Token

brunoocasali:
Create the generate_tenant_token following the specification https://github.com/meilisearch/specifications/pull/89

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -576,6 +576,24 @@ impl Client {
 
         Ok(tasks.results)
     }
+
+    /// Generates a new tenant token.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use meilisearch_sdk::*;
+    /// # futures::executor::block_on(async move {
+    /// # let client = client::Client::new("http://localhost:7700", "masterKey");
+    /// let token = client.generate_tenant_token(serde_json::json!(["*"]), None, None).unwrap();
+    /// let client = client::Client::new("http://localhost:7700", token);
+    /// # });
+    /// ```
+    pub fn generate_tenant_token(&self, search_rules: serde_json::Value, api_key: Option<&str>, expires_at: Option<OffsetDateTime>) -> Result<String, Error> {
+        let api_key = api_key.unwrap_or(&self.api_key);
+
+        crate::tenant_tokens::generate_tenant_token(search_rules, api_key, expires_at)
+    }
 }
 
 #[derive(Deserialize)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -230,6 +230,8 @@ pub mod search;
 pub mod settings;
 /// Module representing the [tasks::Task]s.
 pub mod tasks;
+/// Module that generates tenant tokens.
+mod tenant_tokens;
 
 #[cfg(feature = "sync")]
 pub(crate) type Rc<T> = std::sync::Arc<T>;
--- a/src/search.rs
+++ b/src/search.rs
@@ -283,7 +283,7 @@ mod tests {
     use crate::{client::*, search::*};
     use meilisearch_test_macro::meilisearch_test;
     use serde::{Deserialize, Serialize};
-    use serde_json::{Map, Value};
+    use serde_json::{Map, Value, json};
 
     #[derive(Debug, Serialize, Deserialize, PartialEq)]
     struct Document {
@@ -569,6 +569,41 @@ mod tests {
         query.with_query("harry \"of Fire\"");
         let results: SearchResults<Document> = index.execute_query(&query).await?;
         assert_eq!(results.hits.len(), 1);
+        Ok(())
+    }
+
+    #[meilisearch_test]
+    async fn test_generate_tenant_token_from_client(client: Client, index: Index) -> Result<(), Error> {
+        use crate::key::{KeyBuilder, Action};
+
+        setup_test_index(&client, &index).await?;
+
+        let key = KeyBuilder::new("key for generate_tenant_token test")
+            .with_action(Action::All)
+            .with_index("*")
+            .create(&client).await.unwrap();
+        let allowed_client = Client::new("http://localhost:7700", key.key);
+
+        let search_rules = vec![
+            json!({ "*": {}}),
+            json!({ "*": Value::Null }),
+            json!(["*"]),
+            json!({ "*": { "filter": "kind = text" } }),
+            json!([index.uid.to_string()]),
+        ];
+
+        for rules in search_rules {
+            let token = allowed_client.generate_tenant_token(rules, None, None).expect("Cannot generate tenant token.");
+            let new_client = Client::new("http://localhost:7700", token);
+
+            let result: SearchResults<Document> = new_client.index(index.uid.to_string())
+                .search()
+                .execute()
+                .await?;
+
+            assert!(!result.hits.is_empty());
+        }
+
         Ok(())
     }
 }
--- a/src/tenant_tokens.rs
+++ b/src/tenant_tokens.rs
@@ -1,0 +1,127 @@
+use crate::{
+    errors::* 
+};
+use serde::{Serialize, Deserialize};
+use jsonwebtoken::{encode, Header, EncodingKey};
+use time::{OffsetDateTime};
+use serde_json::Value;
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")] 
+struct TenantTokenClaim {
+    api_key_prefix: String,
+    search_rules: Value,
+    #[serde(with = "time::serde::timestamp::option")]
+    exp: Option<OffsetDateTime>,
+}
+
+pub fn generate_tenant_token(search_rules: Value, api_key: impl AsRef<str>, expires_at: Option<OffsetDateTime>) -> Result<String, Error> {
+    if api_key.as_ref().chars().count() < 8 {
+        return Err(Error::TenantTokensInvalidApiKey)
+    }
+
+    if expires_at.map_or(false, |expires_at| OffsetDateTime::now_utc() > expires_at) {
+        return Err(Error::TenantTokensExpiredSignature)
+    }
+
+    let key_prefix = api_key.as_ref().chars().take(8).collect();
+    let claims = TenantTokenClaim {
+        api_key_prefix: key_prefix,
+        exp: expires_at,
+        search_rules
+    };
+
+    let token = encode(
+        &Header::default(),
+        &claims,
+        &EncodingKey::from_secret(api_key.as_ref().as_bytes()),
+    );
+
+    Ok(token?)
+}
+
+#[cfg(test)]
+mod tests {
+    use serde_json::json;
+    use crate::tenant_tokens::*;
+    use jsonwebtoken::{decode, DecodingKey, Validation, Algorithm};
+    use std::collections::HashSet;
+
+    const SEARCH_RULES: [&str; 1] = ["*"];
+    const VALID_KEY: &str = "a19b6ec84ee31324efa560cd1f7e6939";
+
+    fn build_validation() -> Validation {
+        let mut validation = Validation::new(Algorithm::HS256);
+        validation.validate_exp = false;
+        validation.required_spec_claims = HashSet::new();
+
+        validation
+    }
+
+    #[test]
+    fn test_generate_token_with_given_key() {
+        let token = generate_tenant_token(json!(SEARCH_RULES), VALID_KEY, None).unwrap();
+
+        let valid_key = decode::<TenantTokenClaim>(
+            &token, &DecodingKey::from_secret(VALID_KEY.as_ref()), &build_validation()
+        );
+        let invalid_key = decode::<TenantTokenClaim>(
+            &token, &DecodingKey::from_secret("not-the-same-key".as_ref()), &build_validation()
+        );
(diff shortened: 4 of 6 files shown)


brunoocasali:
For the reviewers: this branch is breaking because I'm waiting for #261 to be merged :)

src/client.rs
```diff
@@ -583,6 +583,27 @@ impl Client {
...
+    /// # let client = client::Client::new("http://localhost:7700", "masterKey");
+    /// let token = client.generate_tenant_token(serde_json::json!(["*"]), None, None).unwrap();
+    /// let client = client::Client::new("http://localhost:7700", token);
+    /// # });
+    /// ```
+    pub fn generate_tenant_token(&self, search_rules: serde_json::Value, api_key: Option<&str>, expires_at: Option<OffsetDateTime>) -> Result<String, Error> {
+        let key = match api_key {
+            Some(api_key) => api_key,
+            _ => &self.api_key
+        };
```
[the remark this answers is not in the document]
irevoire: And I think we should shadow the name `api_key` so we never reuse the option by mistake

src/tenant_tokens.rs
```diff
@@ -0,0 +1,115 @@
...
+    #[serde(with = "time::serde::timestamp::option")]
+    exp: Option<OffsetDateTime>,
+}
+
+pub fn generate_tenant_token(search_rules: serde_json::Value, api_key: &str, expires_at: Option<OffsetDateTime>) -> Result<String, Error> {
+    if api_key.len() < 8 {
+        return Err(Error::TenantTokensInvalidApiKey)
+    }
+
+    if expires_at.is_some() && OffsetDateTime::now_utc() > expires_at.unwrap() {
```
irevoire, line 22:
```suggestion
    if expires_at.map_or(false, |expires_at| OffsetDateTime::now_utc() > expires_at) {
```

It's more common to see it like that in rust

src/tenant_tokens.rs
```diff
@@ -0,0 +1,115 @@
...
+        return Err(Error::TenantTokensExpiredSignature)
+    }
+
+    let key_prefix = &api_key[..8];
```
irevoire, line 26: Here you're creating a new `&str` going from the first byte to the eighth byte.
If the `api_key` can contain utf-8 this can panic.
```suggestion
    // Unwrap is safe since we ensured before that we had at least 8 characters in the api_key
    let key_prefix = &api_key[..api_key.char_indices().nth(8).unwrap().0]];
```

irevoire: Oh, I just noticed you were doing a `.to_string` right below on the `key_prefix`.
Since you're going to allocate a string whatever happens you could write this simpler version;
```suggestion
    let key_prefix = &api_key.chars().take(8).collect();
```
And remove the `.to_string()` below.

src/tenant_tokens.rs
```diff
@@ -0,0 +1,115 @@
...
+        &claims,
+        &EncodingKey::from_secret(api_key.as_bytes()),
+    );
+
+    Ok(token.unwrap())
+}
+
+#[cfg(test)]
+mod tests {
+    use serde_json::{json};
```
irevoire, line 44:
```suggestion
    use serde_json::json;
```

src/tenant_tokens.rs
```diff
@@ -0,0 +1,115 @@
...
+        search_rules
+    };
+
+    let token = encode(
+        &Header::default(),
+        &claims,
+        &EncodingKey::from_secret(api_key.as_bytes()),
+    );
+
+    Ok(token.unwrap())
```
irevoire, line 39: Is this `unwrap` safe?
Shouldn't we return an error instead of panicking if it fails?

src/tenant_tokens.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use serde_json::{json};
+    use crate::{tenant_tokens::*};
```
irevoire, line 45:
```suggestion
    use crate::tenant_tokens::*;
```

irevoire (commented):
Hey @brunoocasali thanks for your PR :grin: 

I think we should add a test with an api-key containing an utf-8 string with a multi-byte character around the 8th byte just to ensure we don't break our code later.

And maybe also a test that try to get a tenant token from meilisearch to see if we deserialize correctly our date?

src/tenant_tokens.rs
```diff
@@ -0,0 +1,115 @@
...
+use jsonwebtoken::{encode, Header, EncodingKey};
+use time::{OffsetDateTime};
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")] 
+struct TenantTokenClaim {
+    api_key_prefix: String,
+    search_rules: serde_json::Value,
+    #[serde(with = "time::serde::timestamp::option")]
+    exp: Option<OffsetDateTime>,
```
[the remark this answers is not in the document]
brunoocasali: Following the specs: https://github.com/meilisearch/specifications/blob/main/text/0089-tenant-tokens.md#3223-payload-example

```json
{
    "apiKeyPrefix": "...",
    "exp": 1641835850, // An expiration date in seconds from 1970-01-01T00:00:00Z UTC
    "searchRules": { .. }
}
```

```jsonc
// using `time::serde::timestamp::option`:
{
  "apiKeyPrefix": "a19b6ec8",
  "searchRules": [ "*" ],
  "exp": 1650468139
}

// using `time::serde::rfc3339::option`:
{
  "apiKeyPrefix": "a19b6ec8",
  "searchRules": [ "*" ],
  "exp": "2022-04-20T15:21:10.165342Z"
}
```

It seems to me that I should keep using the `time::serde::timestamp::option`, right?

src/tenant_tokens.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+pub fn generate_tenant_token(search_rules: serde_json::Value, api_key: &str, expires_at: Option<OffsetDateTime>) -> Result<String, Error> {
```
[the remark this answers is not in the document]
brunoocasali: I liked your suggestions 

But I had some trouble with this one:

image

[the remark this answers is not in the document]
brunoocasali: Oh now it works, I changed the code based on what you sent before https://github.com/meilisearch/meilisearch-rust/pull/263#discussion_r853071604

src/tenant_tokens.rs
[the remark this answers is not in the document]
irevoire: Ah yes, you're right my bad! 
Keep the timestamp :+1:

src/tenant_tokens.rs
```diff
@@ -0,0 +1,128 @@
...
+    #[meilisearch_test]
+    async fn test_generate_token_with_multi_byte_chars() {
+        let key = "Ëa1ทt9bVcL-vãUทtP3OpXW5qPc%bWH5ทvw09";
+        let token = generate_tenant_token(json!(SEARCH_RULES), key, None).unwrap();
+
+        let decoded = decode::<TenantTokenClaim>(
+            &token, &DecodingKey::from_secret(key.as_ref()), &build_validation()
+        ).expect("Cannot decode the token");
+
+        assert_eq!(decoded.claims.api_key_prefix, "Ëa1ทt9bV");
```
irevoire, line 126: perfect :ok_hand:

src/search.rs
```diff
@@ -571,4 +571,39 @@ mod tests {
...
+
+        let key = KeyBuilder::new("key for generate_tenant_token test")
+            .with_action(Action::All)
+            .with_index("*")
+            .create(&client).await.unwrap();
+        let allowed_client = Client::new("http://localhost:7700", key.key);
+
+        let search_rules = vec![
+            json!({'*': {}}),
+            json!({'*': serde_json::Value::Null }),
```
irevoire, line 589:
```suggestion
            json!({'*': Value::Null }),
```

src/tenant_tokens.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    Ok(token?)
+}
+
+#[cfg(test)]
+mod tests {
+    use serde_json::json;
+    use crate::tenant_tokens::*;
+    use jsonwebtoken::{decode, DecodingKey, Validation, Algorithm};
+    use meilisearch_test_macro::meilisearch_test;
```
irevoire, line 48:
```suggestion
```

All the tests in this file do not concern meilisearch and are not async. Thus can create normal tests.

src/tenant_tokens.rs
```diff
@@ -0,0 +1,128 @@
...
+    fn build_validation() -> Validation {
+        let mut validation = Validation::new(Algorithm::HS256);
+        validation.validate_exp = false;
+        validation.required_spec_claims = HashSet::new();
+
+        validation
+    }
+
+    #[meilisearch_test]
+    async fn test_generate_token_with_given_key() {
```
irevoire, line 63:
```suggestion
    #[test]
   fn test_generate_token_with_given_key() {
```

src/tenant_tokens.rs
```diff
@@ -0,0 +1,128 @@
...
+        let invalid_key = decode::<TenantTokenClaim>(
+            &token, &DecodingKey::from_secret("not-the-same-key".as_ref()), &build_validation()
+        );
+
+        assert!(valid_key.is_ok());
+        assert!(invalid_key.is_err());
+    }
+
+    #[meilisearch_test]
+    async fn test_generate_token_without_key() {
```
irevoire, line 78:
```suggestion
    #[test]
   fn test_generate_token_without_key() {
```

src/tenant_tokens.rs
```diff
@@ -0,0 +1,128 @@
...
+        let key = String::from("");
+        let token = generate_tenant_token(json!(SEARCH_RULES), &key, None);
+
+        assert!(token.is_err());
+    }
+
+    #[meilisearch_test]
+    async fn test_generate_token_with_expiration() {
```
irevoire, line 86:
```suggestion
    #[test]
   fn test_generate_token_with_expiration() {
```

src/tenant_tokens.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        let decoded = decode::<TenantTokenClaim>(
+            &token, &DecodingKey::from_secret(VALID_KEY.as_ref()), &Validation::new(Algorithm::HS256)
+        );
+
+        assert!(decoded.is_ok());
+    }
+
+    #[meilisearch_test]
+    async fn test_generate_token_with_expires_at_in_the_past() {
```
irevoire, line 98:
```suggestion
    #[test]
   fn test_generate_token_with_expires_at_in_the_past() {
```

src/tenant_tokens.rs
```diff
@@ -0,0 +1,128 @@
...
+        let exp = OffsetDateTime::now_utc() - time::Duration::HOUR;
+        let token = generate_tenant_token(json!(SEARCH_RULES), VALID_KEY, Some(exp));
+
+        assert!(token.is_err());
+    }
+
+    #[meilisearch_test]
+    async fn test_generate_token_contains_claims() {
```
irevoire, line 106:
```suggestion
    #[test]
   fn test_generate_token_contains_claims() {
```

src/tenant_tokens.rs
```diff
@@ -0,0 +1,128 @@
...
+        let decoded = decode::<TenantTokenClaim>(
+            &token, &DecodingKey::from_secret(VALID_KEY.as_ref()), &build_validation()
+        ).expect("Cannot decode the token");
+
+        assert_eq!(decoded.claims.api_key_prefix, &VALID_KEY[..8]);
+        assert_eq!(decoded.claims.search_rules, json!(SEARCH_RULES));
+    }
+
+    #[meilisearch_test]
+    async fn test_generate_token_with_multi_byte_chars() {
```
irevoire, line 118:
```suggestion
    #[test]
   fn test_generate_token_with_multi_byte_chars() {
```

irevoire (changes requested):
Hey, I think we’ll be able to merge the PR right after these last changes!

Thank you 🎉

src/tenant_tokens.rs
brunoocasali: Solved in earlier versions :)

src/lib.rs
```diff
@@ -230,6 +230,8 @@ pub mod search;
 pub mod settings;
 /// Module representing the [tasks::Task]s.
 pub mod tasks;
+/// Module responsible to generate Tenant Tokens.
```
bidoubiwa, line 233:
```suggestion
/// Module that generates tenant tokens.
```

src/search.rs
```diff
@@ -571,4 +571,39 @@ mod tests {
...
+            .with_action(Action::All)
+            .with_index("*")
+            .create(&client).await.unwrap();
+        let allowed_client = Client::new("http://localhost:7700", key.key);
+
+        let search_rules = vec![
+            json!({'*': {}}),
+            json!({'*': Value::Null }),
+            json!(['*']),
+            json!({ "*": { "filter": "kind = text" } }),
```
bidoubiwa, line 591: Is the choice of different type of quotes expected?

brunoocasali: No it's not! :p

bidoubiwa (approved):
LGTM 🔥🔥

PR #273, opened 2022-04-25
Ensure nested field support

bidoubiwa:
As nested field search is introduced in v0.27 we want to ensure it is possible to use it


irevoire (approved):
coolos :+1:

PR #274, opened 2022-04-26
Add new search parameters highlightPreTag, highlightPostTag and cropMarker

bidoubiwa:
As per https://github.com/meilisearch/meilisearch/issues/2214 new search parameters are introduced:

- `highlightPreTag`
- `highlightPostTag`
- `cropMarker`

### DIFF
--- a/src/search.rs
+++ b/src/search.rs
@@ -173,7 +173,7 @@ pub struct Query<'a> {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(serialize_with = "serialize_attributes_to_crop_with_wildcard")]
     pub attributes_to_crop: Option<Selectors<&'a [AttributeToCrop<'a>]>>,
-    /// Number of words smart cropped around the matching word.
+    /// Maximum number of words including the matched query term(s) contained in the returned cropped value(s).
     /// See [attributes_to_crop](#structfield.attributes_to_crop).
     ///
     /// Default: `10`
--- a/src/search.rs
+++ b/src/search.rs
@@ -173,18 +173,36 @@ pub struct Query<'a> {
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(serialize_with = "serialize_attributes_to_crop_with_wildcard")]
     pub attributes_to_crop: Option<Selectors<&'a [AttributeToCrop<'a>]>>,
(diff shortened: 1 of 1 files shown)


irevoire (approved):
super coolos 🐢

src/search.rs
```diff
@@ -173,18 +173,36 @@ pub struct Query<'a> {
...
     #[serde(serialize_with = "serialize_with_wildcard")]
     pub attributes_to_highlight: Option<Selectors<&'a [&'a str]>>,
+    /// Tag in front of a highlighted term.
+    /// ex: `<mytag>hello world`
+    ///
+    /// Default: `<em>`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub highlight_pre_tag: Option<&'a str>,
+    /// Tag after the a highlighted term.
+    /// ex: `hello world</ mytag>`
```
brunoocasali, line 201: Suggestion:
```
    /// Tag after the highlighted term.
    /// ex: `hello world</mytag>`
```

src/search.rs
```diff
@@ -547,6 +586,62 @@ mod tests {
         Ok(())
     }
 
+    #[meilisearch_test]
+    async fn test_query_customized_crop_marker(client: Client, index: Index) -> Result<(), Error> {
+        setup_test_index(&client, &index).await?;
+
+        let mut query = Query::new(&index);
+        query.with_query("sed do eiusmod");
```
brunoocasali, line 594: I like this way of declaring builders :D
```
let mut query = Query::new(&index)
  .with_query("sed do eiusmod")
  .with_attributes_to_crop(Selectors::All)
```

src/search.rs
```diff
@@ -173,18 +173,36 @@ pub struct Query<'a> {
...
     /// See [attributes_to_crop](#structfield.attributes_to_crop).
     ///
-    /// Default: `200`
+    /// Default: `10`
     #[serde(skip_serializing_if = "Option::is_none")]
     pub crop_length: Option<usize>,
+    /// Marker at the start and the end of a cropped value.
+    /// ex: `...middle of a crop...`
+    ///
+    /// Default: `...`
```
brunoocasali, line 185: There is a subtle difference in the `…` to `...`
```
/// Default: `…`
```

PR #275, opened 2022-04-26
Fix get document doc tests using the wrong index name

bidoubiwa:
Because `get_documents` and `get_document` used the same `index_name` in their example, I stumbled into a concurrency issue where one deleted the index before the assertions of the second. 

To avoid that every doc-tests uses its own index name. Which was the problem here.

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Meilisearch(MeilisearchError { error_message: "Index `get_documents` not found.", error_code: IndexNotFound, error_type: InvalidRequest, error_link: "https://docs.meilisearch.com/errors#index_not_found" })', src/indexes.rs:23:73
```

### DIFF
--- a/src/indexes.rs
+++ b/src/indexes.rs
@@ -213,7 +213,7 @@ impl Index {
     ///
     /// # futures::executor::block_on(async move {
     /// let client = Client::new("http://localhost:7700", "masterKey");
-    /// let movies = client.index("get_documents");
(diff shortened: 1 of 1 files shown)


irevoire (approved):
Cool, it would be handy to have a macro for the doctest but I’m not sure that’s possible

bidoubiwa:
bors merge

PR #278, opened 2022-04-26
Remove jsonindex

### LINKED ISSUE #187: Add `updatedAt` and `createdAt` to `Index`
When manipulating indexes (get/create/update), the MeiliSearch server returns the following information related to the index:
- `uid`: the uid of the index
- `primaryKey`: the primary key of the index
- `updateAt`: when the index has been updated
- `createdAt`: when the index has been created

This SDK, when manipulating indexes (get/create/update) returns an instance of `Index` which already contains the `primaryKey` and `uid` members.

It means we can currently get the `primaryKey` and the `uid`, but not `updatedAt` and `createdAt`.

TODO:
- [ ] Add `updatedAt` and `createdAt` member in the `Index` class/struct (depending on the language).
- [ ] Update the variable in the code where `primaryKey` is already updated. Ex: when getting the index with `getIndex()`
- [ ] Add tests

⚠️ This issue is generated, it means `primaryKey`, `updatedAt`, and `createdAt` might be named differently in this package (ex: `primary_key` instead). Keep the already existing way of naming in this package to stay idiomatic with the language and the repository.
Also, if you are a maintainer, feel free to add any clarification and instruction about this issue related to the specific languages of the repo.

Related to https://github.com/meilisearch/integration-guides/issues/121

Sorry if this is already partially/completely implemented, feel free to let me know about the state of this issue in the repo.

abhizer:
# Pull Request

## What does this PR do?
Fixes #187 
https://github.com/meilisearch/meilisearch-rust/issues/187

## PR checklist

## Todo List: 
 - [x] Add fields updated_at, created_at and primary_key to Index
 - [x] Make get_raw_index return the raw Value that's returned by Meilisearch
 - [x] Update fetch_info to update Index instead of returning JsonIndex
 - [x] Remove JsonIndex and update everything (like get_raw_index) that relies on it to use Index
 - [x] Tests 

*P.S.: I think this requires a thorough review, as a lot of things have been changed.*

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -953,6 +953,7 @@ mod tests {
     #[meilisearch_test]
     async fn test_fetch_info(mut index: Index) {
         let res = index.fetch_info().await;
+
         assert!(res.is_ok());
         assert!(index.updated_at.is_some());
         assert!(index.created_at.is_some());
--- a/src/client.rs
+++ b/src/client.rs
@@ -129,9 +129,8 @@ impl Client {
     /// ```
     /// If you use it directly from an [Index], you can use the method [Index::fetch_info], which is the equivalent method from an index.
     pub async fn get_raw_index(&self, uid: impl AsRef<str>) -> Result<Value, Error> {
-        let idx = self.index(uid.as_ref());
         request::<(), Value>(
-            &format!("{}/indexes/{}", self.host, idx.uid),
+            &format!("{}/indexes/{}", self.host, uid.as_ref()),
             &self.api_key,
             Method::Get,
             200,
--- a/src/client.rs
+++ b/src/client.rs
@@ -52,11 +52,11 @@ impl Client {
     /// # });
     /// ```
     pub async fn list_all_indexes(&self) -> Result<Vec<Index>, Error> {
-        let mut v = vec![];
-        for index in self.list_all_indexes_raw().await? {
-            v.push(Index::from_value(index, self.clone())?);
-        }
-        Ok(v)
+        self.list_all_indexes_raw()
+            .await?
+            .into_iter()
+            .map(|index| Index::from_value(index, self.clone()))
+            .collect()
     }
 
     /// List all [Index]es and returns as Json.
@@ -593,7 +593,12 @@ impl Client {
     /// let client = client::Client::new("http://localhost:7700", token);
     /// # });
     /// ```
-    pub fn generate_tenant_token(&self, search_rules: serde_json::Value, api_key: Option<&str>, expires_at: Option<OffsetDateTime>) -> Result<String, Error> {
+    pub fn generate_tenant_token(
+        &self,
+        search_rules: serde_json::Value,
+        api_key: Option<&str>,
+        expires_at: Option<OffsetDateTime>,
+    ) -> Result<String, Error> {
         let api_key = api_key.unwrap_or(&self.api_key);
 
         crate::tenant_tokens::generate_tenant_token(search_rules, api_key, expires_at)
@@ -943,13 +948,16 @@ mod tests {
         assert!(all_indexes_raw.len() > 0);
         assert!(all_indexes_raw
             .iter()
-            .any(|idx| idx.get("uid").unwrap().as_str().unwrap() == *index.uid));
+            .any(|idx| idx["uid"] == json!(index.uid.to_string())));
     }
 
     #[meilisearch_test]
     async fn test_fetch_info(mut index: Index) {
-        let index = index.fetch_info().await;
-        assert!(index.is_ok());
+        let res = index.fetch_info().await;
+        assert!(res.is_ok());
+        assert!(index.updated_at.is_some());
+        assert!(index.created_at.is_some());
+        assert!(index.primary_key.is_none());
     }
 
     #[meilisearch_test]
--- a/src/client.rs
+++ b/src/client.rs
@@ -52,16 +52,11 @@ impl Client {
     /// # });
     /// ```
     pub async fn list_all_indexes(&self) -> Result<Vec<Index>, Error> {
-        match self.list_all_indexes_raw().await {
-            Ok(json_indexes) => Ok({
-                let mut indexes = Vec::new();
-                for json_index in json_indexes {
-                    indexes.push(json_index.into_index(self))
-                }
-                indexes
-            }),
-            Err(error) => Err(error),
+        let mut v = vec![];
+        for index in self.list_all_indexes_raw().await? {
+            v.push(Index::from_value(index, self.clone())?);
         }
+        Ok(v)
     }
 
     /// List all [Index]es and returns as Json.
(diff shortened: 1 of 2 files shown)


src/indexes.rs
```diff
@@ -72,14 +52,41 @@ impl JsonIndex {
...
+        struct PrivIndex {
+            uid: String,
+            #[serde(with = "time::serde::rfc3339::option")]
+            updatedAt: Option<OffsetDateTime>,
+            #[serde(with = "time::serde::rfc3339::option")]
+            createdAt: Option<OffsetDateTime>,
+            primaryKey: Option<String>,
+        }
+
+        let i: PrivIndex = serde_json::from_value(v).map_err(Error::ParseError)?;
```
irevoire, line 74: This looks pretty strange to me. I think your index should derive Deserialize (and maybe serialize? I’m not sure 🤔)
And then you should use it directly instead of creating this private structure that is basically an `Index`.

src/indexes.rs
```diff
@@ -596,19 +603,20 @@ impl Index {
...
-            Method::Get,
-            200,
-        )
-        .await
+    pub async fn fetch_info(&mut self) -> Result<(), Error> {
+        let v = self.client.get_raw_index(self.uid.as_ref()).await?;
+        *self = match Index::from_value(v, self.client.clone()) {
+            Ok(x) => x,
+            Err(e) => return Err(e),
+        };
```
irevoire, line 618:
```suggestion
        *self = Index::from_value(v, self.client.clone())?;
```

src/indexes.rs
```diff
@@ -628,8 +636,9 @@ impl Index {
     /// # index.delete().await.unwrap().wait_for_completion(&client, None, None).await.unwrap();
     /// # });
     /// ```
-    pub async fn get_primary_key(&self) -> Result<Option<String>, Error> {
-        Ok(self.fetch_info().await?.primaryKey)
+    pub async fn get_primary_key(&mut self) -> Result<Option<String>, Error> {
+        self.fetch_info().await?;
+        Ok(self.primary_key.to_owned())
```
irevoire, line 641: Now that we actually store the primary key somewhere.
I think it would be better to return a reference to the primary key instead of an owned version and let the user clone it if they need to. What do you think?

```suggestion
    pub async fn get_primary_key(&mut self) -> Result<Option<&str>, Error> {
        self.fetch_info().await?;
        Ok(self.primary_key.as_ref())
```

irevoire (changes requested):
Hey, the PR looks really good!
Thanks a lot for your contribution 🎉

src/indexes.rs
abhizer: Oh yeah, that makes sense.

src/indexes.rs
abhizer: I did this because I couldn't quite figure out how to deal with `client` while deserializing. Do you have any suggestions?

irevoire: Ah yes, you're right, my bad. I'm not a fan of this solution, but it's still better than doing tricks with a hollow client that implements `Default` or an `Option<Client>` just for the only time we need to deserialize it.

Let's keep it like that :+1:

irevoire (dismissed):
@bidoubiwa or @brunoocasali, this PR looks good to me, I think we can merge when one of you have the time to review it 😁

src/client.rs
```diff
@@ -109,10 +104,9 @@ impl Client {
     /// # });
     /// ```
     pub async fn get_index(&self, uid: impl AsRef<str>) -> Result<Index, Error> {
-        match self.get_raw_index(uid).await {
-            Ok(raw_idx) => Ok(raw_idx.into_index(self)),
-            Err(error) => Err(error),
-        }
+        let mut idx = self.index(uid.as_ref());
```
brunoocasali, line 107: @irevoire this needs to be mutable?

src/client.rs
```diff
@@ -109,10 +104,9 @@ impl Client {
...
+        idx.fetch_info().await?;
+        Ok(idx)
```
brunoocasali, line 109: We cannot just return: `idx.fetch_info().await?` ?

src/client.rs
```diff
@@ -129,20 +123,30 @@ impl Client {
...
     /// # index.delete().await.unwrap().wait_for_completion(&client, None, None).await.unwrap();
     /// # });
     /// ```
     /// If you use it directly from an [Index], you can use the method [Index::fetch_info], which is the equivalent method from an index.
-    pub async fn get_raw_index(&self, uid: impl AsRef<str>) -> Result<JsonIndex, Error> {
-        Index::fetch_info(&self.index(uid.as_ref())).await
+    pub async fn get_raw_index(&self, uid: impl AsRef<str>) -> Result<Value, Error> {
+        let idx = self.index(uid.as_ref());
+        request::<(), Value>(
+            &format!("{}/indexes/{}", self.host, idx.uid),
```
brunoocasali, line 134: It is possible to use:

```suggestion
        request::<(), Value>(
            &format!("{}/indexes/{}", self.host, uid.as_ref()),
```

src/client.rs
```diff
@@ -937,17 +946,22 @@ mod tests {
...
     #[meilisearch_test]
-    async fn test_fetch_info(index: Index) {
-        let index = index.fetch_info().await;
-        assert!(index.is_ok());
+    async fn test_fetch_info(mut index: Index) {
+        let res = index.fetch_info().await;
+        assert!(res.is_ok());
+        assert!(index.updated_at.is_some());
+        assert!(index.created_at.is_some());
+        assert!(index.primary_key.is_none());
```
brunoocasali, line 960:
```suggestion
    async fn test_fetch_info(mut index: Index) {
        let res = index.fetch_info().await;

        assert!(res.is_ok());
        assert!(index.updated_at.is_some());
        assert!(index.created_at.is_some());
        assert!(index.primary_key.is_none());
```

src/indexes.rs
```diff
@@ -72,14 +52,41 @@ impl JsonIndex {
...
+    pub created_at: Option<OffsetDateTime>,
+    pub updated_at: Option<OffsetDateTime>,
 }
 
 impl Index {
+    /// Internal Function to create an [Index] from `serde_json::Value` and [Client]
+    pub(crate) fn from_value(v: serde_json::Value, client: Client) -> Result<Index, Error> {
+        #[derive(Deserialize, Debug)]
+        #[allow(non_snake_case)]
+        struct PrivIndex {
```
brunoocasali, line 65: What do you think to use `InternalIndex`, `IndexFromSerde` or something like that?

brunoocasali (commented):
Hi @abhizer thanks a lot for your contribution, it is really appreciated ❤️ 

I left some comments in your PR, just to make sure I understood everything!

src/client.rs
abhizer: Yes, the `Index::fetch_info` method takes a mutable reference to self, as it needs to update self.

src/client.rs
abhizer: No, `fetch_info` returns `Result<(), Error>`, as it fetches the information and updates self, and this method, `Client::get_index` returns `Result<Index, Error>`.

src/client.rs
abhizer: Actually yes!

src/indexes.rs
abhizer: Regarding the name of the struct? `IndexFromSerde` makes sense actually.

brunoocasali: Yes, about the name of the struct :)

abhizer: Done!

brunoocasali (approved):
Thanks for your contribution @abhizer ❤️

irevoire (approved):
Thanks for your contribution once again!

bidoubiwa:
Does this PR introduces breaking changes? @irevoire @abhizer

irevoire:
yes that's super breaking

bidoubiwa:
Should the code-samples be [updated here](https://github.com/meilisearch/meilisearch-rust/blob/main/.code-samples.meilisearch.yaml)?

abhizer:
Went through the code samples, didn't notice anything that'd need to be changed.

bidoubiwa:
bors try

bidoubiwa:
bors merge

bidoubiwa:
Flacky tests see #280 and #281

PR #283, opened 2022-05-05
Improve the test-macro

irevoire:
Before this PR, the test macro was using the name of the function as the name of the index.
That caused a bug in our test suite because, at some point, we had two tests with the same name;
```
test indexes::tests::test_fetch_info ... ok
test client::tests::test_fetch_info ... ok
```
Now, this PR uses the whole path of the test instead of only the name of the function, but since
meilisearch doesn't allow `:` in the index name, it then proceeds to replace all the `::` with one
`-`. Thus for the previous examples the generated index name would be
```
indexes-tests-test_fetch_info
client-tests-test_fetch_info
```
That seems like a good solution since rust forbids having two functions with the same name in the
same module, and it's still easily readable in case we need to debug something in meilisearch from
the index name.

### DIFF
--- a/meilisearch-test-macro/src/lib.rs
+++ b/meilisearch-test-macro/src/lib.rs
@@ -89,9 +89,12 @@ pub fn meilisearch_test(params: TokenStream, input: TokenStream) -> TokenStream 
 
         // Now we do the same for the index name
         if use_name {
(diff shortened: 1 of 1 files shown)


brunoocasali (approved):

brunoocasali:
bors merge

PR #286, opened 2022-05-05
Remove the wasm lib.rs from the manifest

irevoire:
Having a link to the web_app `lib.rs` in the examples was making
it try to compile as a dev dependencies which is a constraint we
don’t want.
I removed it and instead included ALL the examples in our workspace
which means we’re now using THEIR cargo.toml instead of ours and
meilisearch-rust is compiled as a normal dependency for these examples.
I also added a `cargo check` for all the examples / tests / code in the
repo. This should execute faster than the test and thus throw errors faster.

This should unlock #254


brunoocasali (dismissed):
Awesome work @irevoire :D

brunoocasali:
bors merge

brunoocasali:
bors merge

brunoocasali:
bors merge

brunoocasali (approved):

brunoocasali:
bors merge

PR #299, opened 2022-07-11
Update task api for v0.28.0

bidoubiwa:
# Pull Request

Related to: https://github.com/meilisearch/integration-guides/issues/205

## Breaking

- [x]  Actions that create a task in Meilisearch now return a `TaskInfo` and not a `Task` object anymore. (ex: add_documents).
- [x]  `get_tasks` now returns a `TasksResults` where the tasks are contained inside the `results` field.
- [x]  `task_id` in `Task` and`TaskInfo` is now a `u32` instead of a `u64`
- [x]  `index_uid` in `Task` and `TaskInfo` is now an Optionnal.
- [x]   TaskType `documentAddition` is replaced with `DocumentAdditionOrUpdate`
- [x] Add pagination to `get_tasks` based of [this](https://github.com/meilisearch/specifications/pull/115) using `limit`, `from`, `next` metadata in the response.
- [x] Rename these task types:
   - documentPartial -> documentAdditionOrUpdate
   - documentAddition -> documentAdditionOrUpdate
   - clearAll -> documentDeletion
- [x]  Possibility to filter the tasks when using `get_tasks` on `indexUid`, `status` and `type`.

// `get_tasks`
```rust
let tasks = client.get_tasks().await.unwrap()
```

// `get_tasks_with`
```rust
let mut query = tasks::TasksQuery::new(&client);
query.with_index_uid(["get_tasks_with"]);
let tasks = client.get_tasks_with(&query).await.unwrap();
```

// get tasks using `execute`
```rust
let mut query = tasks::TasksQuery::new(&client)
   .with_index_uid(["get_tasks_with"])
   .execute()
   .await
   .unwrap();
```

## Routes changes

- [x] `index.get_task` & `index.wait_for_task` now use `GET /tasks/:task_uid`

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -608,7 +608,7 @@ impl Client {
         .await
     }
 
-    /// Get all tasks from the server.
+    /// Get all tasks with query parameters from the server.
     ///
     /// # Example
     ///
@@ -622,18 +622,52 @@ impl Client {
     /// # let client = client::Client::new(MEILISEARCH_HOST, MEILISEARCH_API_KEY);
     ///
     /// let mut query = tasks::TasksQuery::new(&client);
-    /// query.with_index_uid(["get_tasks"]);
-    /// let tasks = client.get_tasks(&query).await.unwrap();
+    /// query.with_index_uid(["get_tasks_with"]);
+    /// let tasks = client.get_tasks_with(&query).await.unwrap();
+    ///
+    /// assert!(tasks.results.len() > 0);
+    /// # client.index("get_tasks_with").delete().await.unwrap().wait_for_completion(&client, None, None).await.unwrap();
+    /// # });
+    /// ```
+    pub async fn get_tasks_with(
+        &self,
+        tasks_query: &TasksQuery<'_>,
+    ) -> Result<TasksResults, Error> {
+        let tasks = request::<&TasksQuery, TasksResults>(
+            &format!("{}/tasks", self.host),
+            &self.api_key,
+            Method::Get(tasks_query),
+            200,
+        )
+        .await?;
+
+        Ok(tasks)
+    }
+
+    /// Get all tasks from the server.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use meilisearch_sdk::*;
+    /// #
+    /// # let MEILISEARCH_HOST = option_env!("MEILISEARCH_HOST").unwrap_or("http://localhost:7700");
+    /// # let MEILISEARCH_API_KEY = option_env!("MEILISEARCH_API_KEY").unwrap_or("masterKey");
+    /// #
+    /// # futures::executor::block_on(async move {
+    /// # let client = client::Client::new(MEILISEARCH_HOST, MEILISEARCH_API_KEY);
+    ///
+    /// let tasks = client.get_tasks().await.unwrap();
     ///
     /// assert!(tasks.results.len() > 0);
     /// # client.index("get_tasks").delete().await.unwrap().wait_for_completion(&client, None, None).await.unwrap();
     /// # });
     /// ```
-    pub async fn get_tasks(&self, tasks_query: &TasksQuery<'_>) -> Result<TasksResults, Error> {
-        let tasks = request::<&TasksQuery, TasksResults>(
+    pub async fn get_tasks(&self) -> Result<TasksResults, Error> {
+        let tasks = request::<(), TasksResults>(
             &format!("{}/tasks", self.host),
             &self.api_key,
-            Method::Get(tasks_query),
+            Method::Get(()),
             200,
         )
         .await?;
@@ -774,8 +808,14 @@ mod tests {
 
     #[meilisearch_test]
     async fn test_get_tasks(client: Client) {
+        let tasks = client.get_tasks().await.unwrap();
+        assert!(tasks.results.len() >= 2);
+    }
+
+    #[meilisearch_test]
+    async fn test_get_tasks_with_params(client: Client) {
         let query = TasksQuery::new(&client);
-        let tasks = client.get_tasks(&query).await.unwrap();
+        let tasks = client.get_tasks_with(&query).await.unwrap();
         assert!(tasks.results.len() >= 2);
     }
 
--- a/src/client.rs
+++ b/src/client.rs
@@ -4,7 +4,7 @@ use crate::{
     key::{Key, KeyBuilder},
     request::*,
     task_info::TaskInfo,
-    tasks::*,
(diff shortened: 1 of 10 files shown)


src/indexes.rs
```diff
@@ -763,30 +758,20 @@ impl Index {
...
-        #[derive(Deserialize)]
-        struct AllTasks {
-            results: Vec<Task>,
-        }
-
-        Ok(request::<(), AllTasks>(
-            &format!("{}/indexes/{}/tasks", self.client.host, self.uid),
+    pub async fn get_tasks(&self) -> Result<TasksResults, Error> {
+        request::<(), TasksResults>(
+            &format!("{}/tasks", self.client.host),
```
alallema, line 769: I think this method should also have the `indexUid` right?
Something like:
```rust
    &format!("{}/tasks?indexUid={}",
        self.client.host,
        self.uid,
```

src/client.rs
```diff
@@ -3,7 +3,9 @@ use crate::{
     indexes::*,
     key::{Key, KeyBuilder},
     request::*,
-    tasks::{async_sleep, Task},
+    task_info::TaskInfo,
+    tasks::*,
+    utils::async_sleep,
```
alallema, line 8: Why this one is import you don't seem to use it?

src/utils.rs
```diff
@@ -0,0 +1,45 @@
+use std::time::Duration;
+
+#[cfg(not(target_arch = "wasm32"))]
+pub(crate) async fn async_sleep(interval: Duration) {
```
alallema, line 4: Where is this method used? And what is it used for?

alallema (approved):
Well done !!! 🔥

src/client.rs
bidoubiwa: Since it is linked to the wasm mystery, I just copy pasted code that was already present and will look into this once we start digging more into the wasm

src/indexes.rs
bidoubiwa: Thanks a lot

src/utils.rs
bidoubiwa: see comment above

src/indexes.rs
```diff
@@ -763,30 +758,18 @@ impl Index {
...
 
-        Ok(request::<(), AllTasks>(
-            &format!("{}/indexes/{}/tasks", self.client.host, self.uid),
-            &self.client.api_key,
-            Method::Get(()),
-            200,
-        )
-        .await?
-        .results)
+        task.with_index_uid([self.uid.as_str()]);
```
alallema, line 770: Where is define this `with_index_uid`?
I'm not sure to understand, you use the `get_tasks()` method from the client and sort every result?

src/utils.rs
alallema: Wasm mystery!

alallema (commented):
I put a comment I don't understand what is done

src/indexes.rs
alallema: Thank for the explanation @bidoubiwa !!!

alallema (approved):
Amazing !!! LGTM!

alallema (approved):
Well Done! 🎉 
LGTM ✨

src/client.rs
```diff
@@ -426,6 +428,7 @@ impl Client {
     /// # let MEILISEARCH_API_KEY = option_env!("MEILISEARCH_API_KEY").unwrap_or("masterKey");
     /// #
     /// # futures::executor::block_on(async move {
+    ///
```
irevoire, line 431:
```suggestion
```

The users are going to see this blank line

src/client.rs
```diff
@@ -606,6 +608,42 @@ impl Client {
...
+    /// ```
+    /// # use meilisearch_sdk::*;
+    /// #
+    /// # let MEILISEARCH_HOST = option_env!("MEILISEARCH_HOST").unwrap_or("http://localhost:7700");
+    /// # let MEILISEARCH_API_KEY = option_env!("MEILISEARCH_API_KEY").unwrap_or("masterKey");
+    /// #
+    /// # futures::executor::block_on(async move {
+    /// # let client = client::Client::new(MEILISEARCH_HOST, MEILISEARCH_API_KEY);
+    ///
```
irevoire, line 623:
```suggestion
```

src/client.rs
```diff
@@ -618,24 +656,23 @@ impl Client {
     /// #
     /// # futures::executor::block_on(async move {
     /// # let client = client::Client::new(MEILISEARCH_HOST, MEILISEARCH_API_KEY);
+    ///
```
irevoire, line 659:
```suggestion
```

src/search.rs
```diff
@@ -112,7 +112,8 @@ type AttributeToCrop<'a> = (&'a str, Option<usize>);
 ///     .with_query("space")
 ///     .with_offset(42)
 ///     .with_limit(21)
-///     .build(); // you can also execute() instead of build()
+///
+/// let res = query.execute().await?.unwrap()
```
irevoire, line 116: I think you missed a `;`

src/tasks.rs
```diff
@@ -369,32 +391,73 @@ impl AsRef<u64> for Task {
...
+    pub client: &'a Client,
+    // Index uids array to only retrieve the tasks of the indexes.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub index_uid: Option<Vec<&'a str>>,
+    // Statuses array to only retrieve the tasks with these statuses.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub status: Option<Vec<&'a str>>,
+    // Types array to only retrieve the tasks with these [TaskType].
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub r#type: Option<Vec<&'a str>>,
```
irevoire, line 407: We should use `TaskType` instead of `&str` here I guess?

src/tasks.rs
```diff
@@ -369,32 +391,73 @@ impl AsRef<u64> for Task {
...
+    // Maximum number of tasks to return
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub limit: Option<u32>,
+    // The if og the first task uid that should be returned
```
irevoire, line 411:
```suggestion
    // The first task uid that should be returned
```

src/tasks.rs
irevoire: Also if the type is going to be used by the end user (which I guess is the case since it’s marked as `pub`) we should rename it to something easier to type. 
I propose;
- `ty`
- `task_type`

alallema (approved):
🔥  🔥  🔥

Issue #265, opened 2022-04-19
Send documents as jsonlines


irevoire:
**Description**
All the functions that create or update documents should send jsonlines instead of a JSON array.

**Why**
That's faster to process by meilisearch and consumes less memory.

**How**
Currently, the issue is that when we send a POST or PUT request in the SDK, [it's automatically flagged as JSON](https://github.com/meilisearch/meilisearch-rust/blob/main/src/request.rs#L129-L139).
We need to update our internal to support JSON for the settings and jsonlines for the documents.

matthias-wright:
I would like to work on this, is this issue still current?
Can the meilisearch server already handle requests that send jsonlines?

irevoire:
Hey @matthias-wright, yes thanks for your interest, this hasn't been implemented.
Meilisearch can handle jsonlines **IF** you set the good content-type and only on the POST and PUT route 👍

postmeback:
If https://github.com/meilisearch/meilisearch-rust/pull/417 was able to fix this, then please close it.

PR #306, opened 2022-07-18
Send documents as jsonlines for POST and PUT requests

matthias-wright:
This PR addresses #265.

The main obstacle I see is that `serde_json` does not offer a way to serialize data to jsonlines. For example, vectors are automatically serialized to json arrays by `serde_json.to_string()`.
A simple workaround is to iterate over the collection and serialize each item individually. Then we can join them using `\n` as the separator to obtain jsonlines:
```rust
let movies = vec![Movie { id: 1, name: String::from("Life of Pi") },
                  Movie { id: 2, name: String::from("Mad Max") },
                  Movie { id: 3, name: String::from("Moana") }];

let mut jsonlines = movies.iter().map(|x| serde_json::to_string(x).unwrap()).collect::<Vec<String>>().join("\n");
jsonlines.push('\n');
```
However, in the [`request`](https://github.com/meilisearch/meilisearch-rust/blob/6c7370d245265f41480595db26213b600d2c53c5/src/request.rs#L92) function, where the data is serialized, `Serialize` is the only trait bound for the data (`Input`). That means we cannot iterate over the data to obtain jsonlines as in the example above.

To work around this we could create another `request` function (and call it `request_jsonlines` or whatever) which only takes in data that is iterable. Then we would also have to create a new `Method` enum. I did not want to do this because it would entail a lot of duplicate code.
Another option would be to write a function that converts a json array to jsonlines. However, that would probably be less efficient than computing the jsonlines directly (as shown above).
Hence, I decided to adapt the `Method` enum, such that it can hold two different types. A type that is iterable (for jsonlines) and a type that is not necessary iterable (for all other calls to `request`).

Let me know what you think about this solution. I am open to suggestions :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,7 +59,7 @@ Also, the WASM example compilation should be checked:
 
 ```bash
 rustup target add wasm32-unknown-unknown
-cargo check --example web_app --target wasm32-unknown-unknown --features=sync
+cargo check -p web_app --target wasm32-unknown-unknown
 ```
 
 Each PR should pass the tests to be accepted.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "meilisearch-sdk"
-version = "0.17.0"
+version = "0.19.0"
 authors = ["Mubelotix <mubelotix@gmail.com>"]
 edition = "2018"
 description = "Rust wrapper for the Meilisearch API. Meilisearch is a powerful, fast, open-source, easy to use and deploy search engine."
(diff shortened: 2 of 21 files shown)


bidoubiwa:
Hey @matthias-wright 
Unfortunately, until this repo is not compatible with Meilisearch v0.28.0 your tests will not pass. I'll come back to you as soon as it is compatible! 

Very sorry about the delay

bidoubiwa:
Hey @matthias-wright 

The repo is now compatible with v0.28.0! There are a lot (a lot lot) of breaking changes in the code base. 
If you'd like to adapt your contribution with the updated code it would be very cool! If not, I completely understand as the code changed a lot.

Don't hesitate to close this PR if you prefer not!

bidoubiwa:
An alternative was introduced to be able to send document with jsonlines here: #417

PR #309, opened 2022-08-02
Add yaup to format query parameters

bidoubiwa:
In preparation of the v0.28.0 compatibility update with Meilisearch, we added `yaup` to handle the future requirements in query parameter formating. 

Currently it was done by hand, example: 

https://github.com/meilisearch/meilisearch-rust/blob/6f98caff09c2439002ac53307d1748a849c9e600/src/indexes.rs#L312-L334

But, with the new version of Meilisearch a lot of query parameters are added in multiple routes. 
To avoid having to do this tedious task over and over again we are using `yaup` that is maintained by Meilisearch and that formats query parameters in a meilisearch compatible way.

Test are failing because they are done against Meilisearch v0.28 and not Meilisearch v0.27. Prouf of successful tests: 

Screenshot 2022-08-02 at 18 46 06

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 time = { version = "0.3.7", features = ["serde-well-known", "formatting", "parsing"] }
 jsonwebtoken = { version = "8", default-features = false }
+yaup = "0.2.0"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 futures = "0.3"
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -24,7 +24,7 @@ pub enum Error {
     TenantTokensInvalidApiKey,
     /// It is not possible to generate an already expired tenant token.
     TenantTokensExpiredSignature,
-    
+
     /// When jsonwebtoken cannot generate the token successfully.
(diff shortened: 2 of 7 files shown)


brunoocasali (commented):
@bidoubiwa I left two 👀 because I would like to know if you will address/remove them in other PRs, or not :)

Cargo.toml
```diff
@@ -19,6 +19,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 time = { version = "0.3.7", features = ["serde-well-known", "formatting", "parsing"] }
 jsonwebtoken = { version = "8", default-features = false }
+yaup = "0.2.0"
```
irevoire, line 22: Wow such a nice name

src/errors.rs
```diff
@@ -67,6 +69,12 @@ impl From<jsonwebtoken::errors::Error> for Error {
     }
 }
 
+impl From<yaup::Error> for Error {
+    fn from(error: yaup::Error) -> Error {
+        Error::Yaup(error)
+    }
+}
```
irevoire, line 76: Could be cool to add [thiserror](https://docs.rs/thiserror/latest/thiserror/) to the repo in another PR to stop writing this by hand like an animal from the prehistoric era

bidoubiwa: I opened an issue :)

irevoire (approved):

PR #313, opened 2022-08-04
Update keys for v0.28.0

bidoubiwa:
## Changes related to the `keys` management

Related to:
- issue: https://github.com/meilisearch/meilisearch/issues/2442
- issue: https://github.com/meilisearch/meilisearch/issues/2369

All the changes:
- [x] `GET /keys` has pagination metadata, added `limit` (default: 20), `offset` (default: 0), `total`.
- [x] `GET /keys/:uid_or_key` can receive either a `:uid` or `:key` to find the key, not just the `:key` how it used to be.
- [x] `DELETE /keys/:uid_or_key` can receive either a `:uid` or `:key` to find the key, not just the `:key` how it used to be.
- [x] `PATCH /keys/:uid_or_key` can receive either a `:uid` or `:key` to find the key, not just the `:key` how it used to be.
- [x] Add the possibility to specify `:uid` to generate deterministic API keys if `null` Meilisearch generates one. It must be a UUID v4 value.
- [x] Add a `:name` attribute to the resource (can be null).
- [x] Add new values to the actions attribute: `keys.get`, `keys.create`, `keys.update` and `keys.delete`.
- [x] `PATCH /keys/:uid_or_key` can update **only** `name` and `description` fields.
- [ ] `apiKeyPrefix` claim is now named `apiKeyUid` and expects the uid of the signing API key as a value.

## Usage
Breaking usage of `get_keys`

`get_keys` without filtering
```rust
    let client = Client::new(MEILISEARCH_HOST, MEILISEARCH_API_KEY);
    let keys = client.get_keys().await.unwrap();
```

`get_keys` with filtering

```rust
    let client = Client::new(MEILISEARCH_HOST, MEILISEARCH_API_KEY);
    let mut query = KeysQuery::new();
    query.with_limit(1);
    let keys = client.get_keys_with(&query).await.unwrap();
```

Response:
```rust
pub struct KeysResults {
    pub results: Vec<Key>,
    pub limit: u32,
    pub offset: u32,
}
```

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -656,6 +656,46 @@ mod tests {
     };
     use meilisearch_test_macro::meilisearch_test;
     use time::OffsetDateTime;
+    use mockito::mock;
+    use std::mem;
+
+    #[meilisearch_test]
+    async fn test_methods_has_qualified_version_as_header() {
+        let mock_server_url = &mockito::server_url();
+        let path = "/hello";
+        let address = &format!("{}{}", mock_server_url, path);
+        let user_agent = &*qualified_version();
+
+        let assertions = vec![
+            (
+                mock("GET", path).match_header("User-Agent", user_agent).create(),
+                request::<String, ()>(address, "", Method::Get, 200)
+            ),
+            (
+                mock("POST", path).match_header("User-Agent", user_agent).create(),
+                request::<String, ()>(address, "", Method::Post("".to_string()), 200)
+            ),
+            (
+                mock("DELETE", path).match_header("User-Agent", user_agent).create(),
+                request::<String, ()>(address, "", Method::Delete, 200)
+            ),
+            (
+                mock("PUT", path).match_header("User-Agent", user_agent).create(),
+                request::<String, ()>(address, "", Method::Put("".to_string()), 200)
+            ),
+            (
+                mock("PATCH", path).match_header("User-Agent", user_agent).create(),
+                request::<String, ()>(address, "", Method::Patch("".to_string()), 200)
+            )
+        ];
+
+        for (m, req) in assertions {
+            let _ = req.await;
+
+            m.assert();
+            mem::drop(m);
+        }
+    }
 
     #[meilisearch_test]
     async fn test_get_keys(client: Client) {
--- a/src/client.rs
+++ b/src/client.rs
@@ -857,18 +857,23 @@ mod tests {
     }
 
     #[meilisearch_test]
-    async fn test_delete_key(client: Client) {
+    async fn test_delete_key(client: Client, name: String) {
         let mut key = KeyBuilder::new();
-        key.with_name("test_delete_key");
+        key.with_name(&name);
         let key = client.create_key(key).await.unwrap();
 
         client.delete_key(&key).await.unwrap();
-        let keys = client.get_keys().await.unwrap();
+        let keys = KeysQuery::new()
+            .with_limit(10000)
+            .execute(&client)
+            .await
+            .unwrap();
+
         assert!(keys.results.iter().all(|k| k.key != key.key));
     }
 
     #[meilisearch_test]
-    async fn test_error_delete_key(mut client: Client) {
+    async fn test_error_delete_key(mut client: Client, name: String) {
         // ==> accessing a key that does not exist
         let error = client.delete_key("invalid_key").await.unwrap_err();
         assert!(matches!(
@@ -882,7 +887,7 @@ mod tests {
 
         // ==> executing the action without enough right
         let mut key = KeyBuilder::new();
-        key.with_name("test_error_delete_key");
+        key.with_name(&name);
         let key = client.create_key(key).await.unwrap();
 
         let master_key = client.api_key.clone();
@@ -915,18 +920,18 @@ mod tests {
     }
 
     #[meilisearch_test]
-    async fn test_create_key(client: Client, description: String) {
+    async fn test_create_key(client: Client, name: String) {
         let expires_at = OffsetDateTime::now_utc() + time::Duration::HOUR;
         let mut key = KeyBuilder::new();
         key.with_action(Action::DocumentsAdd)
-            .with_name("test_create_key")
+            .with_name(&name)
             .with_expires_at(expires_at.clone())
-            .with_description(&description)
+            .with_description("a description")
             .with_index("*");
         let key = client.create_key(key).await.unwrap();
 
         assert_eq!(key.actions, vec![Action::DocumentsAdd]);
-        assert_eq!(&key.description, &Some(description));
+        assert_eq!(&key.name, &Some(name));
         // We can't compare the two timestamp directly because of some nanoseconds imprecision with the floats
         assert_eq!(
             key.expires_at.unwrap().unix_timestamp(),
@@ -938,7 +943,7 @@ mod tests {
     }
 
     #[meilisearch_test]
-    async fn test_error_create_key(mut client: Client) {
+    async fn test_error_create_key(mut client: Client, name: String) {
         // ==> Invalid index name
         /* TODO: uncomment once meilisearch fix this bug: https://github.com/meilisearch/meilisearch/issues/2158
         let mut key = KeyBuilder::new();
@@ -957,7 +962,7 @@ mod tests {
 
         // ==> executing the action without enough right
         let mut no_right_key = KeyBuilder::new();
-        no_right_key.with_name("test_error_create_key");
+        no_right_key.with_name(&format!("{name}_1"));
         let no_right_key = client.create_key(no_right_key).await.unwrap();
 
         // backup the master key for cleanup at the end of the test
@@ -965,7 +970,7 @@ mod tests {
         client.api_key = Arc::new(no_right_key.key.clone());
 
         let mut key = KeyBuilder::new();
-        key.with_name("test_error_create_key_2");
+        key.with_name(format!("{name}_2"));
         let error = client.create_key(key).await.unwrap_err();
 
         assert!(matches!(
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,7 +1,7 @@
 use crate::{
     errors::*,
     indexes::*,
-    key::{Key, KeyBuilder, KeysQuery, KeysResults},
+    key::{Key, KeyBuilder, KeyUpdater, KeysQuery, KeysResults},
     request::*,
     task_info::TaskInfo,
     tasks::{Task, TasksQuery, TasksResults},
@@ -309,6 +309,40 @@ impl Client {
         }
     }
 
+    /// Get the API [Key]s from Meilisearch with parameters.
+    /// See the [meilisearch documentation](https://docs.meilisearch.com/reference/api/keys.html#get-all-keys).
+    ///
+    /// See also [Client::create_key] and [Client::get_key].
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use meilisearch_sdk::{client::*, errors::Error, key::KeysQuery};
+    /// #
+    /// # let MEILISEARCH_HOST = option_env!("MEILISEARCH_HOST").unwrap_or("http://localhost:7700");
+    /// # let MEILISEARCH_API_KEY = option_env!("MEILISEARCH_API_KEY").unwrap_or("masterKey");
+    /// #
+    /// # futures::executor::block_on(async move {
+    /// let client = Client::new(MEILISEARCH_HOST, MEILISEARCH_API_KEY);
+    /// let mut query = KeysQuery::new();
+    /// query.with_limit(1);
+    /// let keys = client.get_keys_with(&query).await.unwrap();
+    ///
+    /// assert_eq!(keys.results.len(), 1);
+    /// # });
+    /// ```
+    pub async fn get_keys_with(&self, keys_query: &KeysQuery) -> Result<KeysResults, Error> {
+        let keys = request::<&KeysQuery, KeysResults>(
+            &format!("{}/keys", self.host),
+            &self.api_key,
+            Method::Get(keys_query),
+            200,
+        )
+        .await?;
+
+        Ok(keys)
+    }
+
     /// Get the API [Key]s from Meilisearch.
     /// See the [meilisearch documentation](https://docs.meilisearch.com/reference/api/keys.html#get-all-keys).
     ///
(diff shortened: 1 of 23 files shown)


src/key.rs
```diff
@@ -65,177 +68,317 @@ impl Key {
...
     /// ```
-    /// # use meilisearch_sdk::{key::KeyBuilder, key::Action, client::Client};
+    /// # use meilisearch_sdk::{key::KeyBuilder, key::Action, client::Client, key::KeyUpdater};
     /// #
     /// # let MEILISEARCH_HOST = option_env!("MEILISEARCH_HOST").unwrap_or("http://localhost:7700");
     /// # let MEILISEARCH_API_KEY = option_env!("MEILISEARCH_API_KEY").unwrap_or("masterKey");
     /// #
     /// # futures::executor::block_on(async move {
     /// # let client = Client::new(MEILISEARCH_HOST, MEILISEARCH_API_KEY);
     ///
```
bidoubiwa, line 188:
```suggestion
```

src/key.rs
```diff
@@ -65,177 +68,317 @@ impl Key {
...
-    /// key.with_indexes(vec!["test", "movies"]);
-    /// # client.delete_key(key).await.unwrap();
+    /// let description = "My not so little lovely test key".to_string();
+    /// let mut key_update = KeyUpdater::new(new_key)
+    ///     .with_description(&description)
+    ///     .execute(&client)
+    ///     .await
+    ///     .unwrap();
+    ///
+    /// # assert_eq!(key_update.description, Some(description));
```
bidoubiwa, line 201:
```suggestion
    /// assert_eq!(key_update.description, Some(description));
```

src/key.rs
```diff
@@ -65,177 +68,317 @@ impl Key {
...
     ///
-    /// key.with_index("test");
-    /// # client.delete_key(key).await.unwrap();
+    /// let mut key_update = KeyUpdater::new(new_key)
+    ///     .with_name(&name)
+    ///     .execute(&client)
+    ///     .await
+    ///     .unwrap();
+    ///
+    /// # assert_eq!(key_update.name, Some(name));
```
bidoubiwa, line 235:
```suggestion
    /// assert_eq!(key_update.name, Some(name));
```

src/key.rs
```diff
@@ -65,177 +68,317 @@ impl Key {
...
+    /// # let MEILISEARCH_HOST = option_env!("MEILISEARCH_HOST").unwrap_or("http://localhost:7700");
+    /// # let MEILISEARCH_API_KEY = option_env!("MEILISEARCH_API_KEY").unwrap_or("masterKey");
+    /// #
+    /// # futures::executor::block_on(async move {
+    /// # let client = Client::new(MEILISEARCH_HOST, MEILISEARCH_API_KEY);
+    ///  let mut keys = KeysQuery::new()
+    ///   .with_limit(1)
+    ///   .execute(&client).await.unwrap();
+    ///
+    /// # assert_eq!(keys.results.len(), 1);
```
bidoubiwa, line 352:
```suggestion
    /// assert_eq!(keys.results.len(), 1);
```

src/key.rs
```diff
@@ -65,177 +68,317 @@ impl Key {
...
```
bidoubiwa, line 376:
```suggestion
    /// assert_eq!(keys.results.len(), 1);
```

src/key.rs
```diff
@@ -359,14 +519,96 @@ impl KeyBuilder {
...
+    /// #
+    /// # futures::executor::block_on(async move {
+    /// # let client = Client::new(MEILISEARCH_HOST, MEILISEARCH_API_KEY);
+    ///  let description = "My not so little lovely test key".to_string();
+    ///
+    ///  let mut key = KeyBuilder::new()
+    ///   .with_description(&description)
+    ///   .execute(&client).await.unwrap();
+    ///
+    /// # assert_eq!(key.description, Some(description));
```
bidoubiwa, line 548:
```suggestion
    /// assert_eq!(key.description, Some(description));
```

src/key.rs
```diff
@@ -439,12 +682,31 @@ pub enum Action {
     #[serde(rename = "stats.get")]
     StatsGet,
     /// Provides access to the [create dump](https://docs.meilisearch.com/reference/api/dump.md#create-a-dump) endpoint. **Not restricted by `indexes`.**
-    #[serde(rename = "dumps.create")]
+    #[serde(rename = "dumps.execute")]
```
alallema, line 685: Is this on the same PR on purpose?

src/key.rs
```diff
@@ -415,7 +658,7 @@ pub enum Action {
     #[serde(rename = "documents.delete")]
     DocumentsDelete,
     /// Provides access to the [create index](https://docs.meilisearch.com/reference/api/indexes.md#create-an-index) endpoint.
-    #[serde(rename = "indexes.create")]
+    #[serde(rename = "indexes.execute")]
```
alallema, line 661: Is this on the same PR on purpose?

src/key.rs
```diff
@@ -6,20 +6,23 @@ use crate::{client::Client, errors::Error};
 /// Represent a [meilisearch key](https://docs.meilisearch.com/reference/api/keys.html#returned-fields)
 /// You can get a [Key] from the [Client::get_key] method.
 /// Or you can create a [Key] with the [KeyBuilder::create] or [Client::create_key] methods.
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize, Clone)]
```
alallema, line 9: I don't see where you made a copy or use `Clone`?

src/key.rs
```diff
@@ -65,179 +68,323 @@ impl Key {
...
+}
+
+impl AsRef<Key> for Key {
+    fn as_ref(&self) -> &Key {
         self
     }
+}
 
-    /// Update the indexes the [Key] can manage.
+#[derive(Debug, Serialize, Deserialize, Clone)]
```
alallema, line 158: Same for Clone I probably miss something

alallema (commented):
Well Done! I really like the new way to handle the key 🏅 
I just have some questions

src/key.rs
```diff
@@ -259,7 +259,7 @@ impl KeyUpdater {
     /// let mut key_update = KeyUpdater::new(&key.key);
     /// key_update.with_description(&description).execute(&client).await;
     ///
-    /// assert_eq!(key_update.description, Some(description));
+    /// # assert_eq!(key_update.description, Some(description));
```
irevoire, line 262: Why do you want to hide this assert? It shows the update worked well I think?

And it’s pretty common all over the documentation, look at the [push method from the stdlib](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.push)

src/client.rs
```diff
@@ -455,28 +485,29 @@ impl Client {
...
     /// # });
     /// ```
-    pub async fn update_key(&self, key: impl AsRef<Key>) -> Result<Key, Error> {
-        request::<&Key, Key>(
+    pub async fn update_key(&self, key: impl AsRef<KeyUpdater>) -> Result<Key, Error> {
+        request::<&KeyUpdater, Key>(
             &format!("{}/keys/{}", self.host, key.as_ref().key),
             &self.api_key,
-            Method::Patch(key.as_ref()),
+            Method::Patch(key.as_ref()), // name and description
```
irevoire, line 510:
```suggestion
            Method::Patch(key.as_ref()),
```

I don't understand this comment, and it’s stressing me out; maybe we should rephrase it... or maybe we should just remove it?

src/client.rs
```diff
@@ -820,27 +851,27 @@ mod tests {
...
-        assert!(keys.iter().any(
-            |k| k.description != "Default Search API Key (Use it to search from the frontend)"
-        ));
-        assert!(keys.iter().any(
-            |k| k.description != "Default Admin API Key (Use it for all other operations. Caution! Do not use it on a public frontend)"
+        assert!(keys.results.len() >= 2);
+        assert!(keys.results.iter().any(|k| k.description
+            != Some("Default Search API Key (Use it to search from the frontend)".to_string())));
+        assert!(keys.results.iter().any(
+            |k| k.description != Some("Default Admin API Key (Use it for all other operations. Caution! Do not use it on a public frontend)".to_string())
```
irevoire, line 858: Once again, since the keys are now paginated, there is some chance we’ll never find these two keys :cry: 
Sadly, I don’t think we can test anything more than the len.

src/client.rs
```diff
@@ -820,27 +851,27 @@ mod tests {
...
         ));
     }
 
     #[meilisearch_test]
-    async fn test_delete_key(client: Client, description: String) {
-        let key = KeyBuilder::new(description);
+    async fn test_delete_key(client: Client) {
+        let mut key = KeyBuilder::new();
+        key.with_name("test_delete_key");
```
irevoire, line 865:
```suggestion
    async fn test_delete_key(client: Client, name: String) {
        let mut key = KeyBuilder::new();
        key.with_name(&name);
```

We have a cool function to get a unique string.

src/client.rs
```diff
@@ -820,27 +851,27 @@ mod tests {
...
         let key = client.create_key(key).await.unwrap();
 
         client.delete_key(&key).await.unwrap();
         let keys = client.get_keys().await.unwrap();
-        assert!(keys.iter().all(|k| k.key != key.key));
+        assert!(keys.results.iter().all(|k| k.key != key.key));
```
irevoire, line 870: I guess we can’t do this test either...
Or maybe instead of removing all these tests we should set the limit to 10000?

src/client.rs
```diff
@@ -820,27 +851,27 @@ mod tests {
...
     }
 
     #[meilisearch_test]
-    async fn test_error_delete_key(mut client: Client, description: String) {
+    async fn test_error_delete_key(mut client: Client) {
```
irevoire, line 874:
```suggestion
    async fn test_error_delete_key(mut client: Client, name: String) {
```

src/client.rs
```diff
@@ -853,7 +884,8 @@ mod tests {
         ));
 
         // ==> executing the action without enough right
-        let key = KeyBuilder::new(description);
+        let mut key = KeyBuilder::new();
+        key.with_name("test_error_delete_key");
```
irevoire, line 888:
```suggestion
        key.with_name(&name);
```

src/client.rs
```diff
@@ -888,42 +920,31 @@ mod tests {
...
+            .with_name("test_create_key")
             .with_expires_at(expires_at.clone())
+            .with_description(&description)
             .with_index("*");
         let key = client.create_key(key).await.unwrap();
 
         assert_eq!(key.actions, vec![Action::DocumentsAdd]);
-        assert_eq!(key.description, description);
+        assert_eq!(&key.description, &Some(description));
         // We can't compare the two timestamp directly because of some nanoseconds imprecision with the floats
```
irevoire, line 933: Here I *think* we should use the unique string generation for the name of the key and hardcode the description to a random value

src/client.rs
```diff
@@ -888,42 +920,31 @@ mod tests {
...
-            expires_at.unix_timestamp()
-        );
-        assert_eq!(remote_key.indexes, vec!["*".to_string()]);
-
         client.delete_key(key).await.unwrap();
     }
 
     #[meilisearch_test]
-    async fn test_error_create_key(mut client: Client, description: String) {
+    async fn test_error_create_key(mut client: Client) {
```
irevoire, line 944:
```suggestion
    async fn test_error_create_key(mut client: Client, name: String) {
```

src/client.rs
```diff
@@ -938,14 +959,16 @@ mod tests {
         */
 
         // ==> executing the action without enough right
-        let no_right_key = KeyBuilder::new(&description);
+        let mut no_right_key = KeyBuilder::new();
+        no_right_key.with_name("test_error_create_key");
```
irevoire, line 963:
```suggestion
        no_right_key.with_name(&format!("{name}_1"));
```

src/client.rs
```diff
@@ -938,14 +959,16 @@ mod tests {
...
         let no_right_key = client.create_key(no_right_key).await.unwrap();
 
         // backup the master key for cleanup at the end of the test
         let master_client = client.clone();
         client.api_key = Arc::new(no_right_key.key.clone());
 
-        let key = KeyBuilder::new(&description);
+        let mut key = KeyBuilder::new();
+        key.with_name("test_error_create_key_2");
```
irevoire, line 971:
```suggestion
        key.with_name(format!("{name}_2"));
```

src/client.rs
```diff
@@ -963,84 +986,22 @@ mod tests {
...
-        assert_eq!(remote_key.actions, vec![Action::DocumentsAdd]);
-        assert_eq!(remote_key.description, description);
-        // We can't compare the two timestamp directly because of some nanoseconds imprecision with the floats
-        assert_eq!(
-            remote_key.expires_at.unwrap().unix_timestamp(),
-            expires_at.unix_timestamp()
-        );
-        assert_eq!(remote_key.indexes, vec!["*".to_string()]);
+        assert_eq!(key.description, Some(description));
+        assert_eq!(key.name, Some(name));
```
irevoire, line 1000: Same as before, I think we should use the string generation for the name and hardcode the description

src/key.rs
irevoire: I don't know if we use it in this PR. But it could be nice for the users to provide a `Clone` implementation since they can manipulate all these structures.

src/key.rs
irevoire: No, and it’s wrong I think

src/key.rs
```diff
@@ -439,12 +682,31 @@ pub enum Action {
...
+    KeyGet,
+    // Provides access to the [create key](https://docs.meilisearch.com/reference/api/keys.html#create-a-key) endpoint.
+    #[serde(rename = "keys.create")]
+    KeyCreate,
+    // Provides access to the [update key](https://docs.meilisearch.com/reference/api/keys.html#update-a-key) endpoint.
+    #[serde(rename = "keys.update")]
+    KeyUpdate,
+    // Provides access to the [delete key](https://docs.meilisearch.com/reference/api/keys.html#delete-a-key) endpoint.
+    #[serde(rename = "keys.delete")]
+    KeyDelete,
```
irevoire, line 704:
```suggestion
    /// Provides access to the [get Key](https://docs.meilisearch.com/reference/api/keys.html#get-one-key) and [get Keys](https://docs.meilisearch.com/reference/api/keys.html#get-all-keys) endpoints.
    #[serde(rename = "keys.get")]
    KeyGet,
    /// Provides access to the [create key](https://docs.meilisearch.com/reference/api/keys.html#create-a-key) endpoint.
    #[serde(rename = "keys.create")]
    KeyCreate,
    /// Provides access to the [update key](https://docs.meilisearch.com/reference/api/keys.html#update-a-key) endpoint.
    #[serde(rename = "keys.update")]
    KeyUpdate,
    /// Provides access to the [delete key](https://docs.meilisearch.com/reference/api/keys.html#delete-a-key) endpoint.
    #[serde(rename = "keys.delete")]
    KeyDelete,
```

This should be documentation, not comments.

irevoire (changes requested):
I have a lot of concerns about the test; I know it probably won’t cause any issue _right now_ but I’m afraid they’ll become flaky after some time thus, most of my comments are to prevent this from happening.

src/key.rs
bidoubiwa: Yes! For me, since these are the `Actions` that you are able to determine in the `actions` of a `Key` It is more a change to the Key API than the dump api.

src/key.rs
bidoubiwa: See this [comment](https://github.com/meilisearch/meilisearch-rust/pull/313#discussion_r953894117)

bidoubiwa: Indeed these are incorrect!

src/client.rs
bidoubiwa: woops

src/key.rs
bidoubiwa: Its to give the possibility to the user to copy if they'd like to

irevoire (approved):

PR #315, opened 2022-08-22
Update indexes for v0.28.0

bidoubiwa:
Related to issue: https://github.com/meilisearch/meilisearch/issues/2373

## All the changes:

- [x] Remove the name field in the index.
- [x] Wrap the indexes objects in a results key.
- [x] GET /indexes has pagination metadata, added limit (default: 20), offset (default: 0), total.

###  Breaking changes

- `client.get_indexes` now returns a `IndexesResults` instead of a `Vec<Index>`
- `client.get_indexes_raw` now returns a `Value` instead of a `Vec<Value>`
- `Index` structure: 
   - `uid` is now a `string` and not a `Arc<string>` and is public
   - `client` is now public
   - `primary_key` is now public

### New

- `client.get_indexes_with` takes as argument `IndexesQuery` giving you the possibility to set `limit` and `offset`.
- `client.get_indexes_raw_with` takes as argument `IndexesQuery` giving you the possibility to set `limit` and `offset`.
- `IndexUpdater` struct gives you the possibility to create an instance of all the fields you would like to change in an Index and then execute the update: `IndexUpdater::new(&client).with_primary_key("my_id".to_string()).execute().await)
- `IndexesQuery` struct allows you to define the filtering applied during the `get_all_indexes_with` call.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -99,7 +99,6 @@ impl Client {
     /// let indexes: IndexesResults = client.list_all_indexes_with(&query).await.unwrap();
     ///
     /// assert_eq!(indexes.limit, 1);
-    /// // TODO: ALIAS
     /// # });
     /// ```
     pub async fn list_all_indexes_with(
@@ -304,9 +303,25 @@ impl Client {
         self.list_all_indexes().await
     }
 
+    /// Alias for [Client::list_all_indexes_with].
+    pub async fn get_indexes_with(
+        &self,
+        indexes_query: &IndexesQuery<'_>,
+    ) -> Result<IndexesResults, Error> {
+        self.list_all_indexes_with(&indexes_query).await
+    }
+
     /// Alias for [Client::list_all_indexes_raw].
     pub async fn get_indexes_raw(&self) -> Result<Value, Error> {
         self.list_all_indexes_raw().await
+    }
+
+    /// Alias for [Client::list_all_indexes_raw_with].
+    pub async fn get_indexes_raw_with(
+        &self,
+        indexes_query: &IndexesQuery<'_>,
+    ) -> Result<Value, Error> {
+        self.list_all_indexes_raw_with(&indexes_query).await
     }
 
     /// Get stats of all indexes.
@@ -1133,7 +1148,7 @@ mod tests {
     #[meilisearch_test]
     async fn test_list_all_indexes(client: Client, index: Index) {
         let all_indexes = client.list_all_indexes().await.unwrap();
-        // TODO: Check total, limit, offset
+
         assert_eq!(all_indexes.limit, 20);
         assert_eq!(all_indexes.offset, 0);
         assert!(all_indexes.results.iter().any(|idx| idx.uid == index.uid));
--- a/src/client.rs
+++ b/src/client.rs
@@ -967,7 +967,6 @@ mod tests {
     #[meilisearch_test]
 
     async fn test_error_delete_key(mut client: Client, name: String) {
-
         // ==> accessing a key that does not exist
         let error = client.delete_key("invalid_key").await.unwrap_err();
         assert!(matches!(
@@ -984,7 +983,6 @@ mod tests {
 
         key.with_name(&name);
         let key = client.create_key(key).await.unwrap();
-
         let master_key = client.api_key.clone();
         // this key has no right
         client.api_key = Arc::new(key.key.clone());
--- a/src/client.rs
+++ b/src/client.rs
@@ -656,6 +656,46 @@ mod tests {
     };
     use meilisearch_test_macro::meilisearch_test;
     use time::OffsetDateTime;
+    use mockito::mock;
+    use std::mem;
+
+    #[meilisearch_test]
+    async fn test_methods_has_qualified_version_as_header() {
+        let mock_server_url = &mockito::server_url();
+        let path = "/hello";
+        let address = &format!("{}{}", mock_server_url, path);
+        let user_agent = &*qualified_version();
+
+        let assertions = vec![
+            (
+                mock("GET", path).match_header("User-Agent", user_agent).create(),
+                request::<String, ()>(address, "", Method::Get, 200)
+            ),
+            (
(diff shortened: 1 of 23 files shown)


src/client.rs
```diff
@@ -1058,17 +1148,38 @@ mod tests {
     #[meilisearch_test]
     async fn test_list_all_indexes(client: Client, index: Index) {
         let all_indexes = client.list_all_indexes().await.unwrap();
-        assert!(all_indexes.len() > 0);
-        assert!(all_indexes.iter().any(|idx| idx.uid == index.uid));
+
+        assert_eq!(all_indexes.limit, 20);
+        assert_eq!(all_indexes.offset, 0);
+        assert!(all_indexes.results.iter().any(|idx| idx.uid == index.uid));
```
irevoire, line 1154: Actually, this can fail now.
If there is more than 20 index in the meilisearch instance, we may not find our own index :unamused: 

I guess we have no choice but to remove this check + the `index` parameter.

src/client.rs
```diff
@@ -1058,17 +1148,38 @@ mod tests {
...
+    #[meilisearch_test]
+    async fn test_list_all_indexes_raw(client: Client) {
         let all_indexes_raw = client.list_all_indexes_raw().await.unwrap();
-        assert!(all_indexes_raw.len() > 0);
-        assert!(all_indexes_raw
-            .iter()
-            .any(|idx| idx["uid"] == json!(index.uid.to_string())));
+
+        assert_eq!(all_indexes_raw["limit"].as_u64().unwrap() as u32, 20);
+        assert_eq!(all_indexes_raw["offset"].as_u64().unwrap() as u32, 0);
```
irevoire, line 1172:
```suggestion
        assert_eq!(all_indexes_raw["limit"], json!(20));
        assert_eq!(all_indexes_raw["offset"], json!(0));
```

let the json do its thing

src/client.rs
```diff
@@ -1058,17 +1148,38 @@ mod tests {
...
+    }
+
+    #[meilisearch_test]
+    async fn test_list_all_indexes_raw_with_params(client: Client) {
+        let mut query = IndexesQuery::new(&client);
+        query.with_limit(1);
+        let all_indexes_raw = client.list_all_indexes_raw_with(&query).await.unwrap();
+
+        assert_eq!(all_indexes_raw["limit"].as_u64().unwrap() as u32, 1);
+        assert_eq!(all_indexes_raw["offset"].as_u64().unwrap() as u32, 0);
```
irevoire, line 1182:
```suggestion
        assert_eq!(all_indexes_raw["limit"], json!(1));
        assert_eq!(all_indexes_raw["offset"], json!(0));
```

Same

irevoire (changes requested):
Mostly noice
image

src/client.rs
```diff
@@ -41,7 +41,23 @@ impl Client {
         }
     }
 
-    /// List all [Index]es and returns values as instances of [Index].
+    fn parse_indexes_results_from_value(&self, value: Value) -> Result<IndexesResults, Error> {
+        let raw_indexes = value["results"].as_array().unwrap();
+
+        let indexes_results = IndexesResults {
+            limit: value["limit"].as_u64().unwrap() as u32,
```
alallema, line 48: Why we don't use `as_u32()` method directly?

src/client.rs
```diff
@@ -1058,17 +1148,38 @@ mod tests {
...
+
+    #[meilisearch_test]
+    async fn test_list_all_indexes_raw(client: Client) {
         let all_indexes_raw = client.list_all_indexes_raw().await.unwrap();
-        assert!(all_indexes_raw.len() > 0);
-        assert!(all_indexes_raw
-            .iter()
-            .any(|idx| idx["uid"] == json!(index.uid.to_string())));
+
+        assert_eq!(all_indexes_raw["limit"].as_u64().unwrap() as u32, 20);
```
alallema, line 1171: Same here about `as_u32()` why don't use this method it doesn't work?

src/indexes.rs
```diff
@@ -47,27 +46,39 @@ use time::OffsetDateTime;
...
+#[serde(rename_all = "camelCase")]
 pub struct Index {
-    pub(crate) uid: Arc<String>,
-    pub(crate) client: Client,
-    pub(crate) primary_key: Option<String>,
-    pub created_at: Option<OffsetDateTime>,
+    #[serde(skip_serializing)]
+    pub client: Client,
+    pub uid: String,
+    #[serde(with = "time::serde::rfc3339::option")]
```
alallema, line 63: Oh cool!

src/client.rs
```diff
@@ -158,13 +238,7 @@ impl Client {
...
     /// Create a corresponding object of an [Index] without any check or doing an HTTP call.
     pub fn index(&self, uid: impl Into<String>) -> Index {
-        Index {
-            uid: Arc::new(uid.into()),
-            client: self.clone(),
-            primary_key: None,
-            created_at: None,
-            updated_at: None,
-        }
+        Index::new(uid, self.clone())
```
alallema, line 241: It is the `Clone` I ask for the Key handler?

alallema (commented):
Great @bidoubiwa just a few questions here ✨

src/client.rs
bidoubiwa: The method `as_u32()` does not exists :(

src/client.rs
bidoubiwa: @irevoire I'm not sure I remember exactly why we had to do this

src/client.rs
bidoubiwa: It does not exists :(

src/client.rs
irevoire: > It is the Clone I ask for the Key handler?

No, it's the `Client` that's being cloned not the key handler.

We had to do this because our methods take `&self: &Client` as a parameter, and `Index::new` takes a `Client` directly, thus we clone the reference and get a real `Client` back.

irevoire (approved):
coolos

PR #318, opened 2022-08-25
Update tenant token api for v0.28.1

bidoubiwa:
## All Changes
- [x] apiKeyPrefix claim is now named apiKeyUid and expects the uid of the signing API key as a value.

## Breaking Changes

The payload of the tenant token is now using the new `uid` field of a key, instead of the key itself:

Example: 
```rust
 let api_key_uid = "76cf8b87-fd12-4688-ad34-260d930ca4f4".to_string();
 let key = "Ëa1ทt9bVcL-vãUทtP3OpXW5qPc%bWH5ทvw09";
 let token = generate_tenant_token(api_key_uid.clone(), json!(SEARCH_RULES), key, None).unwrap();
```

### DIFF
--- a/src/tenant_tokens.rs
+++ b/src/tenant_tokens.rs
@@ -1,34 +1,41 @@
-use crate::{
-    errors::* 
-};
-use serde::{Serialize, Deserialize};
-use jsonwebtoken::{encode, Header, EncodingKey};
-use time::{OffsetDateTime};
+use crate::errors::*;
+use jsonwebtoken::{encode, EncodingKey, Header};
+use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use time::OffsetDateTime;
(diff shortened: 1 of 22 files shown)


src/tenant_tokens.rs
```diff
@@ -1,34 +1,41 @@
...
 use serde_json::Value;
+use time::OffsetDateTime;
+use uuid::Uuid;
 
 #[derive(Debug, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")] 
-struct TenantTokenClaim {
-    api_key_prefix: String,
+#[serde(rename_all = "camelCase")]
+pub struct TenantTokenClaim {
```
bidoubiwa, line 10: no need to be public

irevoire (approved):

PR #322, opened 2022-08-25
Fix failing test for v0.28.0

bidoubiwa:
Some tests were failing but escaped by eyes during their respective implementation. There are fixed here :)


src/client.rs
```diff
@@ -744,9 +745,6 @@ impl Client {
     /// let mut query = tasks::TasksQuery::new(&client);
     /// query.with_index_uid(["get_tasks_with"]);
     /// let tasks = client.get_tasks_with(&query).await.unwrap();
```
bidoubiwa, line 747: put back deletion of index and assertion

irevoire (approved):

PR #324, opened 2022-08-29
Fix wasm failing build due to getrandom in uuid library

bidoubiwa:
Fixes the wasm fail that was caused by
- the uuid library using the `get_random` dependency
- The Request library of wasm that was not up to date


brunoocasali:
Do you think it is worth keeping the UUID v4 client validation?

bidoubiwa:
Well since wasm is used front end anyways and we don't want to allow users to create token on the front end, it has the benefit of removing this method from the wasm build

Issue #308, opened 2022-07-29
Panics during search with a 'Headers' error


AceofSpades5757:
**Description**

When trying to query an index, Rust panics. I haven't seen any other issues on this so I have to assume it has something to do with my current configuration. I've had this issue in the past and simple used the HTTP API instead.

```rust
use wasm_bindgen_futures::spawn_local;

spawn_local(async move {

    let index = client.index("movies");
    let results: SearchResults<Document> = index.search().with_query(&query).execute().await.unwrap();

});
```

**Expected behavior**

To receive results.

**Current behavior**

Rust panics. This is the console logs from the dev tools.

```
frontend-c442de88d55f5c.js:334 panicked at 'called `Result::unwrap()` on an `Err` value: JsValue(TypeError: Failed to execute 'append' on 'Headers': Invalid name
```

**Environment (please complete the following information):**

- OS: Windows
- Meilisearch version: `v0.25.2`, `v0.27.2`
- meilisearch-rust version: `v0.17.0`
- Frontend Framework: `yew`

bidoubiwa:
This seems to be caused by wasm, @irevoire do you have any idea of the issue?

irevoire:
no idea, I guess it has something to do with these two lines;
https://github.com/meilisearch/meilisearch-rust/blob/main/src/request.rs#L109-L110

You were telling me it’s illegal to update the user-agent in chrome, maybe we should not do it. Or at least ignore the result if it fails.

bidoubiwa:
This seems very plausible! I will look into this :) Thanks for raising the issue @AceofSpades5757

PR #330, opened 2022-08-31
Fix wasm panic when trying to change the user-agent

bidoubiwa:
Fixes: #308 

User-agent can not be changed in chromium. It is considered a security breach. Probably because of that wasm does not accept it either.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,7 +59,7 @@ Also, the WASM example compilation should be checked:
 
 ```bash
 rustup target add wasm32-unknown-unknown
-cargo check --example web_app --target wasm32-unknown-unknown --features=sync
+cargo check -p web_app --target wasm32-unknown-unknown
 ```
 
 Each PR should pass the tests to be accepted.
--- a/src/request.rs
+++ b/src/request.rs
@@ -113,7 +113,9 @@ pub(crate) async fn request<Input: Serialize, Output: DeserializeOwned + 'static
     // The 2 following unwraps should not be able to fail
     let mut mut_url = url.clone().to_string();
     let headers = Headers::new().unwrap();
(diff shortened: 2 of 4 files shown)


src/request.rs
```diff
@@ -109,13 +109,11 @@ pub(crate) async fn request<Input: Serialize, Output: DeserializeOwned + 'static
 
     const CONTENT_TYPE: &str = "Content-Type";
     const JSON: &str = "application/json";
-    let user_agent = qualified_version();
 
     // The 2 following unwraps should not be able to fail
     let mut mut_url = url.clone().to_string();
     let headers = Headers::new().unwrap();
     headers.append("Authorization: Bearer", apikey).unwrap();
-    headers.append("User-Agent", &user_agent).unwrap();
```
brunoocasali, line 118: Can you change this to send this information by using the query param?

brunoocasali (changes requested):

src/request.rs
bidoubiwa: Good idea

CONTRIBUTING.md
```diff
@@ -59,7 +59,7 @@ Also, the WASM example compilation should be checked:
 
 ```bash
 rustup target add wasm32-unknown-unknown
-cargo check --example web_app --target wasm32-unknown-unknown --features=sync
+cargo check -p web_app --target wasm32-unknown-unknown
```
brunoocasali, line 62: This changes are not impactful here?

brunoocasali (approved):
LGTM!

CONTRIBUTING.md
bidoubiwa: No, they needed to be changed back when tamo changed the wasm process but it was forgotten.

Issue #367, opened 2022-10-31
Dependabot does not suggest crates updates


bidoubiwa:
Unless the crates we are using are never updated, which I doubt, dependabot does not create automatic PR's to suggest an update.

We should find out why and fix it

curquiza:
I tried to look for the issue, I don't find any...

Here is the old dependabot file we used in the past in Meilisearch, if it can help. I can confirm it was working and that's why we removed it 😂 
https://github.com/meilisearch/meilisearch/pull/1652/files

bidoubiwa:
Thanks! It may be a good start

curquiza:
Looks like it's outdated since we receive dependabot update now

PR #371, opened 2022-11-09
Fix skip-changelog tag in dependabot file

bidoubiwa:
The tag was set as `skip changelog` and not `skip-changelog` whenever dependabot creates a dump

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,7 @@ updates:
   schedule:
     interval: "monthly"
   labels:
-    - 'skip changelog'
+    - 'skip-changelog'
     - 'dependencies'
   rebase-strategy: disabled
 


curquiza (approved):
I want to die....

curquiza:
Is it related to https://github.com/meilisearch/meilisearch-rust/issues/367?

bidoubiwa:
No, it is related to this automatic PR on a lot of repos: https://github.com/meilisearch/meilisearch-rust/pull/363

There is a typo on the tag name. Which does not stop dependabot from doing its job but results in not having the tag added automatically as it does not exists, [see here](https://github.com/meilisearch/meilisearch-rust/pull/369#issuecomment-1297855493).

bidoubiwa:
bors merge

PR #376, opened 2022-11-10
Fix primary key test using the wrong index name

bidoubiwa:
The documentation test on `get_primary_key` was using `movies` as an index. Which creates conflict with the lib test also using `movies` as an index name.

As per the convention on other documentation tests, we are using the name as the function as index name. This ensures that we do not have two times the same index name. This PR fixes this.


alallema (approved):
LGTM! 🔥

bidoubiwa:
bors merge

PR #377, opened 2022-11-10
Add cancel task for v0.30.0

bidoubiwa:
## Breaking changes

- Changed `TasksQuery` structure to `TasksSearchQuery` when building the query parameters for a `get_tasks` call.

before:
```rust
  let mut query = TasksQuery::new(&client)
      .with_index_uids(["movies"])
      .execute()
```

after:

```rust
  let mut query = TasksSearchQuery::new(&client)
      .with_index_uids(["movies"])
      .execute()
```

- attributes of `TasksSearchQuery` are now private. Meaning you cannot do this anymore:
   ```rust
   let mut query = TasksSearchQuery::new(&client);
   query.limit = 10; // Does not work anymore, limit is private
   ```

## Enhancement

- New `tasksCancelQuery` to structure to build the filters to cancel tasks.

## Todo
- Implementation
    - [x] rename `TasksQuery` to `TasksSearchQuery`
    - [x] Creation of `TasksCancelQuery`
    - [x] Create different implementation of `TasksQuery`:
        - [x] TasksPaginationFilters
        - [x] TasksCancelFilters
    - [x] make `TasksQuery` filters private.
    - [x] `with_from` only available on `TasksSearchQuery`
    - [x] `with_limit` only available on `TasksSearchQuery`
    - [x]  Add `canceledBy` in the task return

- Tests:
    - [x] Add tests on `TasksCancelQuery`

### DIFF
--- a/src/tasks.rs
+++ b/src/tasks.rs
@@ -35,6 +35,9 @@ pub enum TaskType {
     TaskCancelation {
         details: Option<TaskCancelation>,
     },
+    SnapshotCreation {
+        details: Option<SnapshotCreation>,
+    },
 }
 
 #[derive(Debug, Clone, Deserialize)]
@@ -75,6 +78,10 @@ pub struct IndexUpdate {
 pub struct IndexDeletion {
     pub deleted_documents: Option<usize>,
 }
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct SnapshotCreation {}
 
 #[derive(Debug, Clone, Deserialize)]
 #[serde(rename_all = "camelCase")]
@@ -261,6 +268,7 @@ impl Task {
     /// # index.delete().await.unwrap().wait_for_completion(&client, None, None).await.unwrap();
     /// # });
     /// ```
+    #[allow(clippy::result_large_err)] // Since `self` has been consumed, this is not an issue
     pub fn try_make_index(self, client: &Client) -> Result<Index, Self> {
         match self {
             Self::Succeeded {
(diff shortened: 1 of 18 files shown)


src/tasks.rs
```diff
@@ -399,165 +412,211 @@ impl AsRef<u32> for Task {
...
-    }
+impl<'a, T> TasksQuery<'a, T> {
     pub fn with_index_uids<'b>(
         &'b mut self,
-        index_uids: impl IntoIterator<Item = &'a str>,
-    ) -> &'b mut TasksQuery<'a> {
-        self.index_uids = Some(index_uids.into_iter().collect());
+        index_uid: impl IntoIterator<Item = &'a str>,
+    ) -> &'b mut TasksQuery<'a, T> {
+        self.index_uids = Some(index_uid.into_iter().collect());
```
alallema, line 495: Why put them all in the singular?

alallema (commented):
Just few comments

src/tasks.rs
```diff
@@ -85,6 +89,14 @@ pub struct DumpCreation {
     pub dump_uid: Option<String>,
 }
 
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TaskCancelation {
+    pub matched_tasks: usize,
+    pub canceled_tasks: usize,
```
alallema, line 96: I think it's better to use `u32`or `u64` for numbers

src/tasks.rs
```diff
@@ -119,6 +131,7 @@ pub struct SucceededTask {
     pub started_at: OffsetDateTime,
     #[serde(with = "time::serde::rfc3339")]
     pub finished_at: OffsetDateTime,
+    pub canceled_by: Option<usize>,
```
alallema, line 134: Same, I think you should use `u32` or `u64`, I think `usize` or `isize` is more for memory no?

PR #379, opened 2022-11-14
Add possibility to have both query parameters and bodies on http requests

bidoubiwa:
Previously, it was only possible to provide one parameter to any HTTP request method.

For example 
- `Post(something)` something would be resolved as body
- `Get(something)` something would be resolved as query parameter

The issue is raised when we need to use `query parameters` on a `POST` route for example. It was not planned.
With this PR, it is now possible to provide both query parameters and bodies to: POST/PATCH/PUT routes

While technically you can send a body to a `DELETE` requests, I did not add it as it is bad practice.

### DIFF
--- a/src/request.rs
+++ b/src/request.rs
@@ -102,12 +102,12 @@ pub(crate) async fn request<
     };
 
     let status = response.status().as_u16();
-    println!("{:?}", &response);
+
     let mut body = response
         .text()
         .await
         .map_err(|e| crate::errors::Error::HttpError(e.into()))?;
-    println!("{:?}", &body);
+
     if body.is_empty() {
         body = "null".to_string();
     }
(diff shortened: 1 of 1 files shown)


src/request.rs
```diff
@@ -4,19 +4,34 @@ use serde::{de::DeserializeOwned, Serialize};
...
-    Patch(T),
-    Put(T),
-    Delete,
+pub(crate) enum Method<Q, B> {
+    Get { query: Q },
+    Post { query: Q, body: B },
+    Patch { query: Q, body: B },
+    Put { query: Q, body: B },
+    Delete { query: Q },
 }
```
irevoire, line 13: It's really nice that you took the time to convert every variant to the struct format :+1:

src/request.rs
```diff
@@ -194,6 +229,10 @@ fn parse_response<Output: DeserializeOwned>(
     expected_status_code: u16,
     body: String,
 ) -> Result<Output, Error> {
+    println!(
+        " expected status code = {} recieved = {}",
+        expected_status_code, status_code
+    );
```
irevoire, line 235:
```suggestion
```

Oops, this shouldn't be here I think

irevoire (changes requested):

irevoire (approved):
noice

bidoubiwa:
bors merge

PR #381, opened 2022-11-15
Add delete tasks for v0.30.0

bidoubiwa:
See [spec](https://github.com/meilisearch/specifications/pull/198)

## Enhancement

It is now possible to delete your tasks: 

example:
```rust
    let mut query = tasks::TasksDeleteQuery::new(&client);
    query.with_index_uids(["movies"]);
    
    let res = client.delete_tasks_with(&query).await.unwrap();
```

This will remove all tasks on the index uid `movies`

## TODO

- [x] Create `TaskDeletion` detail
- [x] Create `TasksDeleteFilters` structure
- [x] Create `TasksDeleteQuery` structure
- [x] Create implementation of `TasksQuery` with `TasksDeleteFilters`
- [x] Add tests

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -4,7 +4,7 @@ use crate::{
     key::{Key, KeyBuilder, KeyUpdater, KeysQuery, KeysResults},
     request::*,
     task_info::TaskInfo,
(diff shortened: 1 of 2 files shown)


alallema (approved):
LGTM! 🍗

PR #382, opened 2022-11-15
Add swap indexes for v0.30.0

bidoubiwa:
As [per the specification](https://github.com/meilisearch/specifications/pull/192)

## Enhancement

It is now possible to swap two indexes

Usage: 

```rust
    client.swap_indexes([&SwapIndexes {
       indexes: (
            "index_a".to_string(),
             "index_b".to_string(),
       ),
    }])
```

Complete Example:

```rust
async fn test_swapping_two_indexes(client: Client) {
        let index_1 = client.index("test_swapping_two_indexes_1");
        let index_2 = client.index("test_swapping_two_indexes_2");

        let t0 = index_1
            .add_documents(
                &[Document {
                    id: "1".to_string(),
                }],
                None,
            )
            .await
            .unwrap();

        index_2
            .add_documents(
                &[Document {
                    id: "2".to_string(),
                }],
                None,
            )
            .await
            .unwrap();

        t0.wait_for_completion(&client, None, None).await.unwrap();

        let task = client
            .swap_indexes([&SwapIndexes {
                indexes: (
                    "test_swapping_two_indexes_1".to_string(),
                    "test_swapping_two_indexes_2".to_string(),
                ),
            }])
            .await
            .unwrap();
        task.wait_for_completion(&client, None, None).await.unwrap();

        let document = index_1.get_document("2").await.unwrap();

        assert_eq!(
            Document {
                id: "2".to_string()
            },
            document
        );
```

## todo

- [x] create `SwapIndexes` struct 
- [x] implement `swap_indexes` on client
- [x] add tests

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -19,9 +19,9 @@ pub struct Client {
     pub(crate) api_key: String,
 }
 
-#[derive(Debug, Clone, Serialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct SwapIndexes {
-    pub indexes: Vec<String>,
+    pub indexes: (String, String),
 }
 
 impl Client {
@@ -364,7 +364,7 @@ impl Client {
         &self,
         indexes: impl IntoIterator<Item = &SwapIndexes>,
     ) -> Result<TaskInfo, Error> {
-        request::<(), &SwapIndexes, TaskInfo>(
+        request::<(), Vec<&SwapIndexes>, TaskInfo>(
             &format!("{}/swap-indexes", self.host),
             &self.api_key,
             Method::Post {
--- a/src/client.rs
+++ b/src/client.rs
@@ -352,12 +352,20 @@ impl Client {
     /// let task_index_2 = client.create_index("swap_index_2", None).await.unwrap();
     ///
     /// // Wait for the task to complete
-    /// let task = task_index_2.wait_for_completion(&client, None, None).await.unwrap();
+    /// task_index_2.wait_for_completion(&client, None, None).await.unwrap();
     ///
-    /// // const swap_indexes = SwapIndexes::new([ SwapIndexes{}])
+    /// let task = client
+    /// .swap_indexes([&SwapIndexes {
+    ///     indexes: (
+    ///         "test_swapping_two_indexes_1".to_string(),
+    ///         "test_swapping_two_indexes_2".to_string(),
+    ///     ),
+    /// }])
+    /// .await
+    /// .unwrap();
     ///
-    /// assert_eq!(index.as_ref(), "create_index");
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -329,6 +334,56 @@ impl Client {
         self.list_all_indexes_raw_with(indexes_query).await
     }
 
+    /// Swaps a list of [Index] tuples.
```
alallema, line 337: Maybe we should specified that is only 2 indexes that will be swapped. Because I already find it confusing that you can give a list

src/client.rs
```diff
@@ -329,6 +334,56 @@ impl Client {
...
+    /// let task_index_2 = client.create_index("swap_index_2", None).await.unwrap();
+    ///
+    /// // Wait for the task to complete
+    /// task_index_2.wait_for_completion(&client, None, None).await.unwrap();
+    ///
+    /// let task = client
+    /// .swap_indexes([&SwapIndexes {
+    ///     indexes: (
+    ///         "test_swapping_two_indexes_1".to_string(),
+    ///         "test_swapping_two_indexes_2".to_string(),
```
alallema, line 361: Is it should be the real name of the index created above "swap_index_1" and "swap_index_2"?

src/client.rs
```diff
@@ -329,6 +334,56 @@ impl Client {
...
+    ///     ),
+    /// }])
+    /// .await
+    /// .unwrap();
+    ///
+    /// # client.index("task_index_1").delete().await.unwrap().wait_for_completion(&client, None, None).await.unwrap();
+    /// # client.index("task_index_2").delete().await.unwrap().wait_for_completion(&client, None, None).await.unwrap();
```
alallema, line 368: Same question why is not the name of the index created: "swap_index_1"?

src/client.rs
```diff
@@ -962,6 +1017,59 @@ mod tests {
...
+        let document = index_1.get_document("2").await.unwrap();
+
+        assert_eq!(
+            Document {
+                id: "2".to_string()
+            },
+            document
+        );
+    }
+
```
alallema, line 1072: Oh, nice test! I will do the same probably

alallema (commented):
Just few question but well done 🔥 🔥 🔥

src/client.rs
bidoubiwa: Since it can be 

```
[
 { indexes: ["a", "b"] },
 { indexes: ["c", "d"] } 
]
```

Its a list of multiple indexes swaps. So in our case a list (as long as you want) of tuples (which means only two)

src/client.rs
bidoubiwa: Ah yes!

alallema (approved):
LGTM! 🎉 🍣

PR #390, opened 2022-11-16
Add `error` field in all types of task status for v0.30.0

bidoubiwa:
## Breaking change

- The error field is always present in any kind of Tasks (SuccededTask, etc..), but its value can be `Null`


alallema (approved):
LGTM but but Clippy seems to say that something is wrong 😅

PR #375, opened 2022-11-09
Add tasks filters for v0.30.0

bidoubiwa:
Add new filters on the `get /tasks` route.

Present in this [spec](https://github.com/meilisearch/specifications/pull/195)

TODO: 
V0.30.0rc1
- TasksQuery: 
   - [x] `index_uid` renamed to `index_uids`
   - [x] `type` renamed to  `types`
   - [x] `status` renamed to `statuses`
   - [x] `uid` renamed to `uids`
- In TaskQuery impl constructor:    
    - [x] `with_index_uid` renamed to `with_index_uids`
    - [x] `with_type` renamed to `with_types`
    - [x] `with_status` renamed to `with_statuses`
    - [x] `with_uid` renamed to `with_uids`
- ErrorsCodes: 
  - [x] invalid_task_types_filter
  - [x] invalid_task_statuses_filter
  - [x] invalid_task_canceled_by_filter
  - [x] invalid_task_uids_filter
  - [x] invalid_task_date_filter 
  - [x] missing_task_filter

V0.30.0rc0
in TasksQuery struct:
  - [x] Add `uid`
  - [x] Add `before_enqueued_at`
  - [x] Add `after_enqueued_at`
  - [x] Add `before_started_at`
  - [x] Add `after_started_at`
  - [x] Add `before_finished_at`
  - [x] Add `after_finished_at`

In TaskQuery impl constructor: 
  - [x] Add `uids: None`
  - [x] Add `before_enqueued_at: None`
  - [x] Add `after_enqueued_at: None`
  - [x] Add `before_started_at: None`
  - [x] Add `after_started_at: None`
  - [x] Add `before_finished_at: None`
  - [x] Add `after_finished_at: None`

in TasksQuery implementation:  
- [x] Add `with_uids()`
- [x] Add `with_before_enqueued_at()`
- [x] Add `with_after_enqueued_at()`
- [x] Add `with_before_started_at()`
- [x] Add `with_after_started_at()`
- [x] Add `with_before_finished_at()`
- [x] Add `with_after_finished_at()`

Tests:
  - [x] Add test on filters

PR #398, opened 2022-11-28
Add canceled by task filter for v0.30.0

bidoubiwa:
Forgot to add this builder function and task filter see #375


alallema (approved):
LGTM! 👾

PR #410, opened 2022-12-22
Fix order of crate publishing

bidoubiwa:
The macro crate should be published before the meilisearch-rust crate since `meilisearch-rust` needs the macro crate to work


bidoubiwa:
bors merge

PR #417, opened 2023-02-05
Provide two methods to add documents from an async reader

irevoire:
# Pull Request

When using Meilisearch-rust to forward documents from another source it’s a pain to use.
These two simple methods let you send raw payload to meilisearch. But you **need** to specify your content-type.

----

## Note

I didn’t implements these method for the wasm version of the crate because `wasm-bindgen-futures` don’t provide an `AsyncReader` trait.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ meilisearch-index-setting-macro = { path = "meilisearch-index-setting-macro", ve
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 futures = "0.3"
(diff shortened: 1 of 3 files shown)


bidoubiwa (commented):
Looks good to me. For the clippy issue I think you just need to rebase with main

bidoubiwa:
bors try

bidoubiwa:
bors try

bidoubiwa:
bors try

bidoubiwa (approved):
bors merge

PR #420, opened 2023-02-06
Avoid the index settings macro from being published

bidoubiwa:
Since it's purpose is only to be used locally, we do not need to publish again.
It removed an additional useless step during the publishing of the meilisearch crate.


bidoubiwa:
Closed as wrong solution to issue

PR #422, opened 2023-02-06
Add script to update the version of meilisearch-index-setting-macro

bidoubiwa:
When publishing the package, both meilisearch and meilisearch-index-setting-macro are published. 
If `meilisearch-index-setting-macro` has not seen its version updated, the publishing will fail as the current version on crates is the same as the one being published.

To tackle this issue without to much hassle, both the macro and `meilisearch-rust` are going to share the same version.

This PR creates a script to automatically update the versions and check if they are correct


alallema (approved):
LGTM! 🚀

bidoubiwa:
bors try

bidoubiwa:
bors merge

PR #423, opened 2023-02-06
Remove unused devdependency in macro

bidoubiwa:
It is unused and causes a circular dependency issue when trying to publish


irevoire (approved):
bors merge

PR #445, opened 2023-03-15
Extend Document trait with function to get index

### LINKED ISSUE #444: Extend Document trait with function to get index
**Description**
Add a `index()` function to the `Document` trait to make it easier to get access to the index.

**Basic example**
Right now, something that implments `Into<String>` needs to be provided when one wants to get access to an index:
```rs
let movie_index = client.index("movies");
```

With this change one could instead write:
```rs
let movie_index = Movie::index(&client);
```
(Provided `Movie` implements the `Document` trait)
This way accessing an index becomes less prone to typos and one would'nt have to keep track of the name of the index, only the associated struct.

**Other**
It might also make sense to add a constant `INDEX_STR` to the trait.
This would have the following benefits:
- it could be used to provide a default implementation for the proposed `Document::index` function
- it could be used when implementing `Document::generate_index` by hand

Adding this to the derive macro would also not require much effort, as the index name is already being extracted there.

I will provide a draft merge request with these changes to act as a base for further discussion.

amaihoefner:
# Pull Request

## Related issue
Fixes #444

## What does this PR do?
- add constant `Document::INDEX_STR` to store name of the index
- add `Document::index` function to provide easier and safer access to index

## PR checklist
- [ x] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [ x] Have you read the contributing guidelines?
- [ x] Have you made sure that the title is accurate and descriptive of the changes?


bidoubiwa (changes requested):
See #444

examples/web_app/src/document.rs
```diff
@@ -13,7 +13,7 @@ pub struct Crate {
     version: String,
 }
 
-// Implement the Document trait so that we can use our struct with Meilisearch
+// Implement the IndexConfig trait so that we can use our struct with Meilisearch
```
bidoubiwa, line 16: I think this documentation may be outdated. We do not implement the indexConfig trait here.

src/dumps.rs
```diff
@@ -6,7 +6,7 @@
 //!
 //! During a [dump export](Client::create_dump), all [indexes](crate::indexes::Index) of the current instance are exported—together with their documents and settings—and saved as a single `.dump` file.
 //!
-//! During a dump import, all indexes contained in the indicated `.dump` file are imported along with their associated [documents](crate::documents::Document) and [settings](crate::settings::Settings).
+//! During a dump import, all indexes contained in the indicated `.dump` file are imported along with their associated [index_configs](crate::document::IndexConfig) and [settings](crate::settings::Settings).
```
bidoubiwa, line 9:
```suggestion
//! During a dump import, all indexes contained in the indicated `.dump` file are imported along with their associated documents and [settings](crate::settings::Settings).
```

bidoubiwa (changes requested):

bidoubiwa:
Only thing missing is updating the title and the body of the PR. After that it can be released :)

bidoubiwa:
bors merge

bidoubiwa (approved):
Thanks a lot 🙏

PR #448, opened 2023-03-20
Fix the build issues introduced by latest serde patch

bidoubiwa:
In the latest serde patch version a breaking change led to a build error. 
By installing the project for the first time, the latest patch versions are installed. This is not the case in an already existing project unless `cargo update` is ran.


irevoire (approved):
bors merge

PR #459, opened 2023-04-04
Feat trait to make clients customizable

### LINKED ISSUE #452: Let users customize their web client
Following: https://github.com/meilisearch/meilisearch-rust/pull/426#issuecomment-1466375016

It would be nice to make the web client customizable, i.e:
- Add a feature flag to get rid of `ishac` (by default, we should continue to use it)
- Add a `new_with_client` function to create a [`Client`](https://github.com/meilisearch/meilisearch-rust/blob/9b0eafe32101ea9e8670de4c1407bfe6e411b0e8/src/client.rs#L17-L21) with anything that implements [`request`](https://github.com/meilisearch/meilisearch-rust/blob/9b0eafe32101ea9e8670de4c1407bfe6e411b0e8/src/request.rs#L118-L210)

shimatar0:
# Pull Request

## Related issue
Fixes #452 

## What does this PR do?
- This is a WIP state and I'll add it next time.
- There are still many problems to be solved. I will comment on the code.

## PR checklist
- [x] Fix the settings macro
- [x] Fix compiling without default features


src/tasks.rs
```diff
@@ -185,7 +186,7 @@ impl AsRef<u32> for EnqueuedTask {
 
 #[derive(Debug, Clone, Deserialize)]
 #[serde(rename_all = "camelCase", tag = "status")]
-pub enum Task {
+pub enum Task<Http: HttpClient> {
```
shimatar0, line 189: compiler error
```
error[E0392]: parameter `Http` is never used
   --> src/tasks.rs:189:15
    |
189 | pub enum Task<Http: HttpClient> {
    |               ^^^^ unused parameter
    |
    = help: consider removing `Http`, referring to it in a field, or using a marker such as `PhantomData`
```

However, Task feels that this parameter is necessary to call Client
https://github.com/meilisearch/meilisearch-rust/blob/8a4f62d49a2433a1db6d1e06426fb27731403ce5/src/tasks.rs#L263

shimatar0:
Sorry. It's not in working condition at all yet. It may be a more difficult task than think. 
It is very likely that this PR will only share the situation and not finalize the merge.

src/tasks.rs
shimatar0: Similar compile errors have been reported by others...

src/client.rs
```diff
@@ -46,7 +46,7 @@ impl<Http: HttpClient> Client<Http> {
         Client {
             host: host.into(),
             api_key: api_key.map(|key| key.into()),
-            http_client: Box::new(IsahcClient),
+            http_client: IsahcClinet,
```
shimatar0, line 49: This could be an elementary problem...
It seems to me that you are fulfilling.
```
mismatched types
expected type parameter `Http`
           found struct `request::IsahcClinet`
```

shimatar0:
This sounds like a lot to think about!
Return value needed to be changed to meet async trait.
So it looks like we need to match the `parse_response`.
```
error[E0277]: the trait bound `dyn futures::Future<Output = Result<Box<dyn std::marker::Send>, errors::Error>> + std::marker::Send: Deserialize<'_>` is not satisfied
   --> src/request.rs:123:9
    |
123 |         parse_response(status, expected_status_code, body)
    |         ^^^^^^^^^^^^^^ the trait `Deserialize<'_>` is not implemented for `dyn futures::Future<Output = Result<Box<dyn std::marker::Send>, errors::Error>> + std::marker::Send`
```

src/lib.rs
```diff
@@ -235,7 +235,7 @@ pub mod errors;
 pub mod indexes;
 /// Module containing the [key::Key] struct.
 pub mod key;
-mod request;
+pub mod request;
```
shimatar0, line 238: I've had to expose the request mod because in order to use HttpClient

src/documents.rs
```diff
@@ -38,29 +39,30 @@ use serde::{de::DeserializeOwned, Deserialize, Serialize};
...
 
+use crate::request::HttpClient;
 use crate::settings::Settings;
 use crate::tasks::Task;
 use crate::Client;
 use crate::{errors::Error, indexes::Index};
 
-#[async_trait]
-pub trait IndexConfig {
+#[async_trait(?Send)]
```
shimatar0, line 57: I needed it to build with wasm32.
But this might not be a very good way to write it.

src/dumps.rs
```diff
@@ -70,21 +70,23 @@ impl Client {
...
-            self.get_api_key(),
-            Method::Post {
-                query: (),
-                body: (),
-            },
-            202,
-        )
-        .await
+        self.http_client
+            .clone()
```
shimatar0, line 74: Clone is now required for the call.
And other places.

examples/cli-app-with-reqwest/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "cli-app-with-requwest"
```
irevoire, line 2:
```suggestion
name = "cli-app-with-reqwest"
```

examples/cli-app-with-reqwest/src/main.rs
```diff
@@ -0,0 +1,282 @@
...
+        Output: DeserializeOwned + 'static,
+    >(
+        self,
+        url: &str,
+        apikey: Option<&str>,
+        method: Method<Query, Body>,
+        content_type: &str,
+        expected_status_code: u16,
+    ) -> Result<Output, Error> {
+        unimplemented!("stream_request is not implemented for RequwestClient")
```
irevoire, line 132:
```suggestion
        unimplemented!("stream_request is not implemented for ReqwestClient")
```

src/request.rs
```diff
@@ -1,220 +1,390 @@
...
+        self,
+        url: &str,
+        apikey: Option<&str>,
+        method: Method<Query, Body>,
+        expected_status_code: u16,
+    ) -> Result<Output, Error>
+    where
+        Query: Serialize + Send + Sync,
+        Body: Serialize + Send + Sync,
+        Output: DeserializeOwned + 'static + Send;
```
irevoire, line 42:
```suggestion
    async fn request<Query, Body, Output>(
        self,
        url: &str,
        apikey: Option<&str>,
        method: Method<Query, Body>,
        expected_status_code: u16,
    ) -> Result<Output, Error>
    where
        Query: Serialize + Send + Sync,
        Body: Serialize + Send + Sync,
        Output: DeserializeOwned + 'static + Send;
```

Huum, I’m struggling to see a difference between these two 🤔 
Shouldn’t we be able to simply define it once without the wasm/no-wasm attribute?

src/request.rs
```diff
@@ -1,220 +1,390 @@
...
+        content_type: &str,
+        expected_status_code: u16,
+    ) -> Result<Output, Error>;
+}
 
-    if query.is_empty() {
-        Ok(url.to_string())
-    } else {
-        Ok(format!("{url}?{query}"))
+#[derive(Debug, Clone, Serialize)]
```
irevoire, line 60:
```suggestion
#[derive(Debug, Clone, Copy, Default, Serialize)]
```

src/request.rs
```diff
@@ -1,220 +1,390 @@
...
-    let builder = Request::builder().header(header::USER_AGENT, qualified_version());
-    let builder = match apikey {
-        Some(apikey) => builder.header(header::AUTHORIZATION, format!("Bearer {apikey}")),
-        None => builder,
-    };
+impl Default for IsahcClient {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
irevoire, line 73:
```suggestion
```

Since we can derive `Default` we don’t need this anymore

src/request.rs
```diff
@@ -1,220 +1,390 @@
...
+#[async_trait(?Send)]
+impl HttpClient for IsahcClient {
+    #[cfg(not(target_arch = "wasm32"))]
+    async fn request<Query, Body, Output>(
+        self,
+        url: &str,
+        apikey: Option<&str>,
+        method: Method<Query, Body>,
+        expected_status_code: u16,
+    ) -> Result<Output, Error>
```
irevoire, line 84: Here it would be better to split the two implementations in different block I think;

```rust
// First, the non-wasm impl block where the whole block is gated behind the cfg thingy

#[cfg(not(target_arch = "wasm32"))]
#[async_trait(?Send)]
impl HttpClient for IsahcClient {
async fn request<Query, Body, Output>(
        self,
        url: &str,
       ...) -> Result<Output, Error> { ... }

    async fn stream_request<
        'a,
        Query: Serialize + Send + Sync,
        Body: futures_io::AsyncRead + Send + Sync + 'static,
        Output: DeserializeOwned + 'static,
    >(
        self,
        url: &str,
        ...) -> Result<Output, Error> { ... }
}

// And then the wasm block that doesn’t contains the stream_request method at all

#[cfg(target_arch = "wasm32")]
#[async_trait(?Send)]
impl HttpClient for IsahcClient {
async fn request<Query, Body, Output>(
        self,
        url: &str,
       ...) -> Result<Output, Error> { ... }
}
```

What do you think?

irevoire (changes requested):
You have some warning on your new example

examples/cli-app-with-reqwest/Cargo.toml
shimatar0: Fixed
https://github.com/meilisearch/meilisearch-rust/pull/459/commits/1f722a791adecf46def0234913ae18da85479747

examples/cli-app-with-reqwest/src/main.rs
shimatar0: Fixed
https://github.com/meilisearch/meilisearch-rust/pull/459/commits/1f722a791adecf46def0234913ae18da85479747

src/request.rs
shimatar0: You are right!
I didn't bother to separate them 😂
fixed
https://github.com/meilisearch/meilisearch-rust/pull/459/commits/2a1cc7a8e6b088c15c7c34ac3fe98246c1b96947

src/request.rs
shimatar0: fixed
https://github.com/meilisearch/meilisearch-rust/pull/459/commits/e7e2c355b1e410a02cb81d06cb4d9fa0f6ec9b9c

src/request.rs
shimatar0: fixed
https://github.com/meilisearch/meilisearch-rust/pull/459/commits/e7e2c355b1e410a02cb81d06cb4d9fa0f6ec9b9c

src/request.rs
shimatar0: It certainly doesn't feel right to have methods with the same name in the same block.
I think it is appropriate to separate them by `flag` and keep the necessary ones, so I've modified it!🙏

https://github.com/meilisearch/meilisearch-rust/pull/459/commits/2a1cc7a8e6b088c15c7c34ac3fe98246c1b96947

examples/cli-app-with-reqwest/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[dependencies]
+meilisearch-sdk = {path="../.."}
+futures = "0.3"
+serde = { version="1.0",   features = ["derive"] }
+serde_json = "1.0"
+lazy_static = "1.4.0"
+reqwest = "0.11.16"
+async-trait = "0.1.51"
+tokio = {version = "1.27.0",  features = ["full"]}
```
bidoubiwa, line 17:
```suggestion
tokio = { version = "1.27.0",  features = ["full"] }

```

examples/cli-app-with-reqwest/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "cli-app-with-reqwest"
+version = "0.0.0"
+edition = "2021"
+publish = false
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+meilisearch-sdk = {path="../.."}
```
bidoubiwa, line 10: Here you could the add `no-default-feature` option

examples/cli-app-with-reqwest/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+futures = "0.3"
+serde = { version="1.0",   features = ["derive"] }
```
bidoubiwa, line 12:
```suggestion
serde = { version="1.0", features = ["derive"] }
```

examples/cli-app-with-reqwest/Cargo.toml
bidoubiwa, line 17:
```suggestion
tokio = {version = "1.27.0", features = ["full"]}
```

examples/settings.rs
```diff
@@ -1,14 +1,15 @@
...
 
 // we need an async runtime
 #[tokio::main(flavor = "current_thread")]
 async fn main() {
-    let client: Client = Client::new("http://localhost:7700", Some("masterKey"));
+    let client: Client<IsahcClient> = Client::new("http://localhost:7700", Some("masterKey"));
 
     // We try to create an index called `movies` with a primary_key of `movie_id`.
-    let my_index: Index = client
+    let my_index: Index<IsahcClient> = client
```
bidoubiwa, line 12: This is a huge breaking and impacts the user experience considerably :( 

Can we have the `isahcClient` by default which you only would have to declare if you are in `no default feature` mode?

src/client.rs
```diff
@@ -121,7 +144,7 @@ impl Client {
     /// # movies.delete().await.unwrap().wait_for_completion(&client, None, None).await.unwrap();
     /// # });
     /// ```
-    pub fn multi_search(&self) -> MultiSearchQuery {
+    pub fn multi_search(&self) -> MultiSearchQuery<Http> {
```
bidoubiwa, line 147: Is it possible to hide the http client here so that the user does not have to provide it if they use ishac?

src/request.rs
```diff
@@ -1,220 +1,378 @@
...
-                .body(AsyncBody::from_reader(body))
-                .map_err(|_| crate::errors::Error::InvalidRequest)?
-                .send_async()
-                .await?
+
+        parse_response(status, expected_status_code, body)
+    }
+}
+
+#[cfg(feature = "isahc")]
```
bidoubiwa, line 258: This can be removed, isahc does not work in wasm

src/request.rs
```diff
@@ -1,220 +1,378 @@
...
+#[cfg(target_arch = "wasm32")]
+#[async_trait(?Send)]
+impl HttpClient for IsahcClient {
```
bidoubiwa, line 261: not sure this work

examples/cli-app-with-reqwest/Cargo.toml
shimatar0: fixed
https://github.com/meilisearch/meilisearch-rust/pull/459/commits/f7986c7031912da09b09fdbc07366f6b668dcfa1

examples/cli-app-with-reqwest/Cargo.toml
shimatar0: fixed
https://github.com/meilisearch/meilisearch-rust/pull/459/commits/daa85e8835be2b272bb6f2cd7d66e733648e7c8b

examples/cli-app-with-reqwest/Cargo.toml
shimatar0: changed
https://github.com/meilisearch/meilisearch-rust/pull/459/commits/35221ea4d4d3850a5d59a169dd2757441439c803

examples/settings.rs
shimatar0: You are right. Thanks for checking and taking the time to help me out 😊
This was a destructive change for users who had specified the type :(

I modified the SDK so that users who use it by default can use it without worrying about traits.(but a little forced😢)
However, I'm not sure if this is a good thing to do, as it also requires changes to the export and to `lib.rs`. 

All in all, this is a big change 
so I hope the maintainer team will consider the decision and any better response 🙏

https://github.com/meilisearch/meilisearch-rust/pull/459/commits/6d96c9f956b6ef02e6a92b41b017d5a06eb840e0

bidoubiwa: What do you think @irevoire ?

irevoire: Yeah, I don't see any other way to do it tbh.
What would you think of having a `prelude` like they have in some libs @bidoubiwa?
For example, `rayon` uses one (look at their examples. They all starts with `use rayon::prelude::*;` instead of picking each thing you need separately: https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.for_each_with
serenity also exposes a prelude: https://docs.rs/serenity/latest/serenity/client/struct.Client.html

I feel like that's the easiest way for us to define all the types we need to overwrite in a prelude file gated by the feature being used.

That being said, it's a « big » breaking change, it'll probably break a lot of code, but the fix for the end users will be way easier => get rids of all your imports and just write `use meilisearch_sdk::prelude::*;`
The cool thing for us is that we can easily change the type alias in the prelude file without being breaking in the future + create a prelude for the wasm + a prelude for isahc

src/request.rs
shimatar0: Fixed to use `WebSysClient` for wasm.

https://github.com/meilisearch/meilisearch-rust/commit/71f2abebb91da8e9b3afe29745730804f7f869ec

src/request.rs
shimatar0: Removed this

https://github.com/meilisearch/meilisearch-rust/commit/71f2abebb91da8e9b3afe29745730804f7f869ec

examples/settings.rs
shimatar0: I tried to modify the form to use `prelude`!
Please let me know if I'm wrong🙏

https://github.com/meilisearch/meilisearch-rust/pull/459/commits/bd04a9bd5be75cdd03b43b0dc0ed92b4c6bbe199
https://github.com/meilisearch/meilisearch-rust/pull/459/commits/5b326278139e521261c659a1a0e83a84890cb1ae

src/client.rs
shimatar0: I have thought about it a bit, but it seems difficult due to the presence of the Client in the `MultiSearchQuery` structure🤔
If the use case of `MultiSearchQuery` is only to be called as a method from the `Client` structure, I don't think there is that much of a problem. 
However, if you are going to implement `MultiSearchQuery` directly, I think it would be better to modify it.

Cargo.toml
```diff
@@ -38,8 +38,8 @@ wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
 
 [features]
-default = ["isahc-static-curl"]
-isahc-static-curl = ["isahc/static-curl"]
+default = ["isahc", "isahc-static-curl"]
```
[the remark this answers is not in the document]
shimatar0: Added description to readme.
I'm a little unsure whether to write a code example or not in here, 
because writing something that moves is going to be huge!🤔

https://github.com/meilisearch/meilisearch-rust/pull/459/commits/5993c50552fa5f7a4ca63edf5dc3657b6800ce9b

novacrazy:
What is the status of this? 

I'm looking into using Meilisearch, but having to bundle an entire copy of curl into my binary just because of this issue is giving me hesitation. I use hyper/reqwest extensively within my server already.

irevoire:
The pr is mainly finished. It should go over a last round of review, but with the MeiliSearch release coming out this Monday, we didn't find the time to merge it.
I hope we’ll be able to merge it next week

bidoubiwa (changes requested):
Hey! @shimatar0. Sorry for the delay.

Thanks a lot again for the great work. These are the changes we would like to see in your PR:

- Add all the public types in the preludes (DocumentsQuery, Settings, etc)
- Update the readme using the `prelude` import.
- Add a section in the readme mentioning the custom HTTP feature, with a link pointing to your cli app with reqwest.
- Update the code-samples `.code-samples.meilisearch.yml`

gibbz00:
Is there anything I can do to get this PR past the finishing line? I'm interested in adding a `reqwest` client backend but that contribution would be blocked by this.

examples/cli-app-with-reqwest/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        let status = response.status().as_u16();
+
+        let mut body = response.text().await.unwrap();
+
+        if body.is_empty() {
+            body = "null".to_string();
+        }
+
+        parse_response(status, expected_status_code, &body, url.to_string())
+        // parse_response(status, expected_status_code, body)
```
irevoire, line 108:
```suggestion
```

src/client.rs
irevoire: Done

src/client.rs
```diff
@@ -159,18 +217,19 @@ impl Client {
     ///
     /// ```
     /// # use meilisearch_sdk::{client::*, indexes::*};
-    /// #
+    ///
```
irevoire, line 220:
```suggestion
    /// #
```

irevoire (changes requested):

PR #460, opened 2023-04-05
Doc comment standardization

### LINKED ISSUE #372: Add documentation on builder methods.
Currently some builder methods across the code base do not have any documentation. See for example: 

https://github.com/meilisearch/meilisearch-rust/blob/1e687c78d7d14e255474380db99b6b13941b8682/src/search.rs#L287-L303

It would be better if each of these methods are explained and showcase an example. As for example here

https://github.com/meilisearch/meilisearch-rust/blob/1e687c78d7d14e255474380db99b6b13941b8682/src/indexes.rs#L1239-L1279

IsaacCloos:
# Pull Request
This work addresses a slew of inconsistencies I've noticed in this projects documentation. These were found while trying to establish a basis to abide to for new doc comments coming soon to resolve #372 

## Related issue
Partially addresses #372 

## What does this PR do?
- hide client instantiation from `example` view. It was 50/50 throughout the project so I push the 'hide' strategy. I personally believe it cleans up the method examples, but if you'd like to have it visible everywhere instead I can totally make that change 😄. I just think it would be beneficial to have some consistency here for users. (making the docs more predictable and all)
- removed a few ```` ```rust ```` annotations (not the code!!). Per the [docs](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#documentation-tests) stating `rust` is the default language.
- made assertions visible where possible in the docs. This was 50/50 throughout and I believe that rounding out an example with a `print` or `assert` command is a better user experience.
- implemented clippy recommendations in every existing doc comment in the project. Not **ALL** suggestions, but most. Lots of tab corrections, which I think make the comments much nicer to read.
- fixed spacing inconsistencies in assert(s). E.g. `.len()>0` changed to `.len() > 0`
- bolded `Default` declarations (a trend I've noticed in other libraries).
- did my best to split up paragraphs of text where I believe the intention was for the text to be separated but markdown puts new lines together if they are just one line return apart. Some were using the backtick method, but I changed them all to be double returns.
- added a handful of missing references: [Key], [Index], ext
- hid additional references making examples more concise
- hid certain missed comments and closures that resulted in jarring or inaccurate examples

## PR checklist
- [X] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [X] Have you read the contributing guidelines?
- [X] Have you made sure that the title is accurate and descriptive of the changes?

# Criticism Welcome
I thought being a little opinionated about this first pass at a general "clean up" could prove fruitful if it sparked discussions about what YOU want to see in this project. I'd be more than happy to make any necessary changes to get this the way you'd like. If you have any questions about decisions I made, please let me know! I'm new to Rust and always trying to understand the dos and don'ts better 😄 

Thanks for the feedback! 

p.s. I am aware that this PR is everything BUT what the linked ticket is requesting 😅. I call it 'Preliminary work' 😆

### DIFF
--- a/examples/cli-app/src/main.rs
+++ b/examples/cli-app/src/main.rs
@@ -67,13 +67,13 @@ async fn build_index() {
     // serialize the string to clothes objects
     let clothes: Vec<Clothes> = serde_json::from_str(content).unwrap();
 
-    //create displayed attributes
+    // create displayed attributes
     let displayed_attributes = ["article", "cost", "size", "pattern"];
 
     // Create ranking rules
     let ranking_rules = ["words", "typo", "attribute", "exactness", "cost:asc"];
 
-    //create searchable attributes
+    // create searchable attributes
     let searchable_attributes = ["seaon", "article", "size", "pattern"];
 
     // create the synonyms hashmap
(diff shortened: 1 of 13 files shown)


bidoubiwa:
bors merge

PR #472, opened 2023-05-22
Add delete_documents_with method for Meilisearch v1.2

bidoubiwa:
following this spec: https://github.com/meilisearch/specifications/pull/236

- Add a method `index.delete_documents_with`  that lets you select a range of documents to delete based on the provided filters.

### DIFF
--- a/src/documents.rs
+++ b/src/documents.rs
@@ -420,21 +420,17 @@ mod tests {
         setup_test_index(&client, &index).await?;
         index
             .set_filterable_attributes(["id"])
-            .await
-            .unwrap()
+            .await?
             .wait_for_completion(&client, None, None)
-            .await
-            .unwrap();
+            .await?;
+
         let mut query = DocumentDeletionQuery::new(&index);
         query.with_filter("id = 1");
-
         index
             .delete_documents_with(&query)
-            .await
-            .unwrap()
+            .await?
             .wait_for_completion(&client, None, None)
-            .await
-            .unwrap();
+            .await?;
         let document_result = index.get_document::<MyObject>("1").await;
 
         match document_result {
@@ -446,6 +442,35 @@ mod tests {
                 _ => panic!("The error was expected to be a Meilisearch error, but it was not."),
             },
         }
+
+        Ok(())
+    }
+
+    #[meilisearch_test]
+    async fn test_delete_documents_with_filter_not_filterable(
+        client: Client,
+        index: Index,
+    ) -> Result<(), Error> {
+        setup_test_index(&client, &index).await?;
+
+        let mut query = DocumentDeletionQuery::new(&index);
+        query.with_filter("id = 1");
+        let error = index
+            .delete_documents_with(&query)
+            .await?
+            .wait_for_completion(&client, None, None)
+            .await?;
+
+        let error = error.unwrap_failure();
+
+        assert!(matches!(
+            error,
+            MeilisearchError {
+                error_code: ErrorCode::InvalidDocumentFilter,
+                error_type: ErrorType::InvalidRequest,
+                ..
+            }
+        ));
 
         Ok(())
     }
--- a/src/documents.rs
+++ b/src/documents.rs
@@ -310,16 +310,19 @@ pub struct DocumentDeletionQuery<'a> {
     /// Filters to apply.
     ///
     /// Read the [dedicated guide](https://docs.meilisearch.com/reference/features/filtering.html) to learn the syntax.
-    pub filter: &'a str,
+    pub filter: Option<&'a str>,
 }
 
 impl<'a> DocumentDeletionQuery<'a> {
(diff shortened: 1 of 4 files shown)


src/tasks.rs
```diff
@@ -66,6 +66,7 @@ pub struct DocumentAdditionOrUpdate {
 pub struct DocumentDeletion {
     pub provided_ids: Option<usize>,
     pub deleted_documents: Option<usize>,
+    pub original_filter: String,
```
irevoire, line 69: The normal document deletion shouldn’t work anymore with your code.
This is the type meilisearch can return:

```suggestion
    pub provided_ids: usize,
    pub deleted_documents: Option<usize>,
    pub original_filter: Option<String>,
```

src/documents.rs
```diff
@@ -387,6 +412,41 @@ mod tests {
...
+        query.with_filter("id = 1");
+
+        index
+            .delete_documents_with(&query)
+            .await
+            .unwrap()
+            .wait_for_completion(&client, None, None)
+            .await
+            .unwrap();
+        let document_result = index.get_document::<MyObject>("1").await;
```
irevoire, line 435: I think we should try to parse all the error codes that can be returned by this route, I left a few comments on all the things we should add to the test.
Also, I think we can get rid of all the `.unwrap()`.

```suggestion
    async fn test_delete_documents_with(client: Client, index: Index) -> Result<(), Error> {
        setup_test_index(&client, &index).await?;
        
        //We should ensure it doesn't work if we don’t use any parameter in the builder and we can parse the new error code
        let mut query = DocumentDeletionQuery::new(&index);
        index.delete_documents_with(&query) // TODO
        // TODO assert thingy
        
        // Then, just to be sure, it would also be nice to ensure we can handle the error returned by meilisearch in case the field is not filterable
        query.with_filter("id = 1");
         index.delete_documents_with(&query) // TODO
        // TODO assert thingy
        
        index
            .set_filterable_attributes(["id"])
            .await
            .unwrap() // you can use `?` to reduce the size of the test I think 🤔 
            .wait_for_completion(&client, None, None)
            .await
            .unwrap(); // same
        let mut query = DocumentDeletionQuery::new(&index);
        
        index
            .delete_documents_with(&query)
            .await
            .unwrap() // same
            .wait_for_completion(&client, None, None)
            .await
            .unwrap(); // same
        let document_result = index.get_document::<MyObject>("1").await;
        
        // one last thing that we should test since it returns a new error code is that we get the right error code with a bad filter
        query.with_filter("doggo");
```

src/documents.rs
```diff
@@ -301,11 +302,36 @@ impl<'a> DocumentsQuery<'a> {
...
+
+impl<'a> DocumentDeletionQuery<'a> {
+    pub fn new(index: &Index) -> DocumentDeletionQuery {
+        DocumentDeletionQuery { index, filter: "" }
+    }
+
+    pub fn with_filter<'b>(&'b mut self, filter: &'a str) -> &'b mut DocumentDeletionQuery<'a> {
+        self.filter = filter;
+        self
+    }
```
irevoire, line 324: If the filter is absolutely mandatory to make the request work we should get it from the `::new`, right?

```suggestion
    /// Filters to apply.
    ///
    /// Read the [dedicated guide](https://docs.meilisearch.com/reference/features/filtering.html) to learn the syntax.
    pub filter: &'a str,
}

impl<'a> DocumentDeletionQuery<'a> {
    pub fn new(index: &Index, filter: &str) -> DocumentDeletionQuery {
        DocumentDeletionQuery { index, filter }
    }
```

-----

Or maybe the filter should be optional, and we should let meilisearch returns the right error message, i.e.: « empty filter» vs « missing filter »

```suggestion
    /// Filters to apply.
    ///
    /// Read the [dedicated guide](https://docs.meilisearch.com/reference/features/filtering.html) to learn the syntax.
    pub filter: Option<&'a str>,
}

impl<'a> DocumentDeletionQuery<'a> {
    pub fn new(index: &Index) -> DocumentDeletionQuery {
        DocumentDeletionQuery { index, filter: None }
    }

    pub fn with_filter<'b>(&'b mut self, filter: &'a str) -> &'b mut DocumentDeletionQuery<'a> {
        self.filter = Some(filter);
        self
    }
```

irevoire (changes requested):

src/documents.rs
bidoubiwa: Let's go with the second option to take advantage of the errors returned by Meilisearch.

src/tasks.rs
bidoubiwa: `provided_ids` is not returned when calling the [delete all documents route](https://www.meilisearch.com/docs/reference/api/documents#delete-all-documents).

I'll put both in optional

irevoire (approved):

PR #473, opened 2023-05-23
Add get documents filter field for ms v1.2

bidoubiwa:
as per the specification: https://github.com/meilisearch/specifications/pull/234

- Add filter and the builder method `with_filter` on the `DocumentQuery` structure. 

The `filter` field works precisely like the `filter` field used on the `search` method. See [the docs on how to use filters](https://www.meilisearch.com/docs/learn/advanced/filtering#filter-basics).

### DIFF
--- a/src/indexes.rs
+++ b/src/indexes.rs
@@ -1,7 +1,7 @@
 use crate::{
     client::Client,
     documents::{DocumentQuery, DocumentsQuery, DocumentsResults},
-    errors::Error,
+    errors::{Error, MeilisearchCommunicationError, MeilisearchError, MEILISEARCH_VERSION_HINT},
     request::*,
     search::*,
     task_info::TaskInfo,
@@ -466,6 +466,39 @@ impl Index {
         &self,
         documents_query: &DocumentsQuery<'_>,
     ) -> Result<DocumentsResults<T>, Error> {
+        if documents_query.filter.is_some() {
+            let url = format!("{}/indexes/{}/documents/fetch", self.client.host, self.uid);
+            return request::<(), &DocumentsQuery, DocumentsResults<T>>(
+                &url,
+                self.client.get_api_key(),
+                Method::Post {
+                    body: documents_query,
+                    query: (),
+                },
+                200,
+            )
+            .await
+            .map_err(|err| match err {
+                Error::MeilisearchCommunication(error) => {
+                    Error::MeilisearchCommunication(MeilisearchCommunicationError {
+                        status_code: error.status_code,
+                        url: error.url,
+                        message: Some(format!("{}.", MEILISEARCH_VERSION_HINT)),
+                    })
+                }
+                Error::Meilisearch(error) => Error::Meilisearch(MeilisearchError {
+                    error_code: error.error_code,
+                    error_link: error.error_link,
+                    error_type: error.error_type,
+                    error_message: format!(
+                        "{}\n{}.",
+                        error.error_message, MEILISEARCH_VERSION_HINT
+                    ),
+                }),
+                _ => err,
+            });
+        }
+
         let url = format!("{}/indexes/{}/documents", self.client.host, self.uid);
         request::<&DocumentsQuery, (), DocumentsResults<T>>(
             &url,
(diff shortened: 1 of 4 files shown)


src/indexes.rs
```diff
@@ -466,6 +466,30 @@ impl Index {
...
+                },
+                200,
+            )
+            .await
+            .map_err(|err| match err {
+                Error::MeilisearchCommunication(error) => {
+                    Error::MeilisearchCommunication(MeilisearchCommunicationError {
+                        status_code: error.status_code,
+                        url: error.url,
+                        message: Some("It might not be working because you're not up to date with the Meilisearch version that updated the get_documents_with method.".to_string()),
```
alallema, line 486:
```suggestion
                        message: Some("Hint: It might not be working because you're not up to date with the Meilisearch version that updated the get_documents_with method.".to_string()),
```
We don't put the `Hint` here?

src/indexes.rs
```diff
@@ -1,7 +1,7 @@
 use crate::{
     client::Client,
     documents::{DocumentQuery, DocumentsQuery, DocumentsResults},
-    errors::Error,
+    errors::*,
```
alallema, line 4:
```suggestion
    errors::{Error, MeilisearchCommunicationError},
```
I'm not sure it's the case in Rust but it's not better to just import what you need?

alallema (commented):
Well Done 🎉 
It's crazy that there was no MeilisearchCommunicationError until now 😳

src/indexes.rs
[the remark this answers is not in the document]
bidoubiwa: > Wouldn't it be better to create a way to add a message directly to an existing error? I'm not sure how to do it in rust.

There is no initial message with isahc on a server returning a HTTP code above 400. So for example, 405 has no message with `Method Not Allowed` written.

> but if in the next update of Meilisearch, it is possible to add other fields in addition to filters, then the error may be a MeilisearchAPI error right?


Not sure I understand this part. Do you have an example?

In case this answers your question:
The communication error is triggered when:
- `status_code >= 400 `
- it was not able to parse the error into a MeilisearchApiError
- it was able to reach Meilisearch (in contrario of a httpError) 

So if Meilisearch returns an error with `unknown field` because in the future we add new params on this route, then, there is no hint. There are a lot of reason that might cause an error and I rather not confuse the user with hints if Meilisearch already provides an error message.

src/indexes.rs
bidoubiwa: It's a design choice indeed, since there are already a lot of `*` in these imports I left it like that. But in the future we might want to change all `*` to specific imports.

src/indexes.rs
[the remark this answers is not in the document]
bidoubiwa: After discussing it in private, we decided that the hint should only be provided when there is a communication error with meilisearch

bidoubiwa: Changed!

src/indexes.rs
bidoubiwa: changed nonetheless!

bidoubiwa:
I'v added a new test @alallema 

`test_get_documents_with_invalid_filter`

src/request.rs
```diff
@@ -345,16 +351,26 @@ fn parse_response<Output: DeserializeOwned>(
...
+                return Err(Error::MeilisearchCommunication(
+                    MeilisearchCommunicationError {
+                        status_code,
+                        message: None,
+                        url,
+                    },
+                ));
+            }
+            Err(Error::ParseError(e))
+        }
```
alallema, line 373: We've talked about this, but as a reminder, the central issue of the error handler is this:
https://github.com/meilisearch/integration-guides/issues/19
```
Following the previous issue about error handler naming, here are the 3 minimum errors the SDKs should raise:

- MeiliSearchApiError: MeiliSearch (HTTP) sends back a 4xx or a 5xx. There are most of the SDKs errors.
- MeiliSearchCommunicationError: problem with communication, for example, the MeiliSearch instance is unreachable.
- MeiliSearchError: for any other errors in the SDK (wrong parameter if the SDK checks the parameters...etc)
```
I know that for you it should be in the MeilisearchCommunicationError but this is not the case in most SDK. It might be worth opening an issue about this in integration-guide
WDYT? I can do it if you want

alallema (approved):
LGTM! 🎉

PR #478, opened 2023-05-30
Fix json serialization error if meilisearch response contains facet_stats

### LINKED ISSUE #474: Error parsing response JSON for facetStats
**Description**
If response from Mailisearch contains facetStats (facet field must contain numeric values to make it  happen), sdk returns error. It seems it happens because FacetStats struct is defined with whole number field types. But meilisearch response json contains floats. Changing u32 to f64 fixes this problem.
```
pub struct FacetStats {
    pub min: u32,
    pub max: u32,
}
```

**Expected behavior**
No error should happen.

**Current behavior**
Error parsing response JSON returned

**Environment (please complete the following information):**
- OS: Linux
- Meilisearch version: 1.1
- meilisearch-rust version: 0.23.1

Vedmak:
# Pull Request

## Related issue
Fixes #474

## What does this PR do?
- change field types of FacetStats struct from u32 to f64
- add test for facet_stats

## PR checklist

### DIFF
--- a/src/search.rs
+++ b/src/search.rs
@@ -50,8 +50,8 @@ pub struct SearchResult<T> {
 #[derive(Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct FacetStats {
-    pub min: u32,
-    pub max: u32,
+    pub min: f64,
+    pub max: f64,
 }
 #[derive(Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase")]
--- a/src/search.rs
+++ b/src/search.rs
@@ -557,6 +557,7 @@ mod tests {
         id: usize,
         value: String,
         kind: String,
+        number: i32,
         nested: Nested,
     }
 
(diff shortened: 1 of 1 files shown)


src/search.rs
```diff
@@ -50,8 +50,8 @@ pub struct SearchResult<T> {
 #[derive(Deserialize, Debug, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct FacetStats {
-    pub min: u32,
-    pub max: u32,
+    pub min: f64,
+    pub max: f64,
```
bidoubiwa, line 54: does `u64` not fix this issue? I don't think facet stats can return a negative number

bidoubiwa (commented):
Thanks a lot for this contribution 🙏

src/search.rs
sanders41: `u64` is an unsigned integer so according to issue it would still error with this since it needs a decimal.

bidoubiwa: Oh yeah ... my bad 🙈

Vedmak: Looked in meilisearch repo, and there this struct has f64 fields https://github.com/search?q=repo%3Ameilisearch%2Fmeilisearch%20FacetStats&type=code

bidoubiwa (approved):

bidoubiwa:
bors merge

PR #493, opened 2023-07-17
Re-export public structures for better import path

### LINKED ISSUE #246: Better import path
**Description**
It's quite boring to start every file with `use meilisearch_sdk::{document::*, client::*};`
It would be better if in the `lib.rs` we were re-exporting every structure like that:
```rust
pub use client::*;
pub use index::*;
...
```
**Basic example**

And thus let us (=the users) write import like that;
```rust
use meilisearch_sdk::*;
```

Or if you want a few basic structure structures:
```rust
use meilisearch_sdk::{Client, Index, Setting};
// instead of
use meilisearch_sdk::{client::Client, indexes::Index, settings::Setting};
```

**Other**

Also, I never remember which module name is plural or singular so it would hide our inconsistencies in a non-breaking change.

Another solution would be to provide a `prelude` like in most libraries but I'm not sure we really need it currently since our lib is so smol :thinking:

adriantombu:
# Pull Request

## Related issue
Fixes #246

## What does this PR do?
- I think the PR title is already self-explanatory 😄

## PR checklist

### DIFF
--- a/examples/cli-app/src/main.rs
+++ b/examples/cli-app/src/main.rs
@@ -1,6 +1,6 @@
 use futures::executor::block_on;
 use lazy_static::lazy_static;
-use meilisearch_sdk::{client::*, settings::Settings};
+use meilisearch_sdk::{client::*, Settings};
 use serde::{Deserialize, Serialize};
 use std::fmt;
 use std::io::stdin;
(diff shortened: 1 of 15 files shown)


bidoubiwa (approved):
thanks a lot 🙏

bidoubiwa (approved):
bors merge

Issue #541, opened 2024-01-22
[v1.6] support `embedders` setting


CommanderStorm:
**Description**
Currently, the `embedders` cannot be added via the sdk, only via http.

**Basic example**

instead of

```sh
curl \       
  -X PATCH 'http://localhost:7700/indexes/movies/settings' \
  -H 'Content-Type: application/json' \
  --data-binary '{
    "embedders": {
      "default": {
        "source": "huggingFace",
        "model": "bge-base-en-v1.5",
        "documentTemplate": "A movie titled '{{doc.title}}' whose description starts with {{doc.body|truncatewords: 20}}"
      }
    }
  }'
```

```rust
Settings::new().with_embedders(HashSet::from("default", Embedder::HuggingFace{...}))
```

**Other**
If adding such features to the sdk BEFORE they are out of the experimental stage, I could implement this.
If not, that is fine too ^^

CommanderStorm:
@curquiza 
You added the `enhancement` label. Does this mean I can work on this, or what is your stance on experimental features in sdks? ^^

curquiza:
Sorry, you are right, I was not clear at all.
Yes, you can work on this 😊 

I'm curious, do you use this feature + rust SDK, or is it only to make an opensource contribution? 😊

CommanderStorm:
> I'm curious, do you use this feature + rust SDK

Yes, but not yet in production (in my uni, not a company).
I have toyed with it in an experimental project and thought that adding this to the sdk might be a good excuse to learn how this should be configured.

Strift:
Closing this issue as outdated in favor of #646

PR #554, opened 2024-03-02
Vector search embedder

CommanderStorm:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch-rust/issues/541

## What does this PR do?
- introduces the `experimental-vector-search`-feture flag
- Adds the required settings
- adds the `hybrid` field to search via the vector search 

TODO:
- [ ] added `{get,set,reset}_embedders` settings
- [ ] add testcases for these settings
- [ ] find a combination of semantic search model + configuration that does not fail the assumptions (see search testcase) spectacularly
- [ ] double-check that every line is guarded by `cfg`-makros

## PR checklist

### DIFF
--- a/src/search.rs
+++ b/src/search.rs
@@ -135,6 +135,22 @@ pub enum Selectors<T> {
     All,
 }
 
+#[cfg(feature = "experimental-vector-search")]
+#[derive(Debug, Serialize, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct HybridSearch<'a> {
+    /// Indicates one of the embedders configured for the queried index
+    ///
+    /// **Default: `"default"`**
+    embedder: &'a str,
+    /// number between `0` and `1`:
+    /// - `0.0` indicates full keyword search
+    /// - `1.0` indicates full semantic search
+    ///
+    /// **Default: `0.5`**
+    semantic_ratio: f32,
+}
+
 type AttributeToCrop<'a> = (&'a str, Option<usize>);
 
 /// A struct representing a query.
@@ -325,6 +341,12 @@ pub struct SearchQuery<'a> {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub(crate) index_uid: Option<&'a str>,
+
+    /// EXPERIMENTAL
+    /// Defines whether to utilise previously defined embedders for semantic searching
+    #[cfg(feature = "experimental-vector-search")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub hybrid: Option<HybridSearch<'a>>,
 }
 
 #[allow(missing_docs)]
@@ -353,6 +375,8 @@ impl<'a> SearchQuery<'a> {
             show_ranking_score: None,
             matching_strategy: None,
             index_uid: None,
+            #[cfg(feature = "experimental-vector-search")]
+            hybrid: None,
         }
     }
     pub fn with_query<'b>(&'b mut self, query: &'a str) -> &'b mut SearchQuery<'a> {
@@ -525,6 +549,19 @@ impl<'a> SearchQuery<'a> {
         self.index_uid = Some(&self.index.uid);
         self
     }
+    #[cfg(feature = "experimental-vector-search")]
+    pub fn with_hybrid<'b>(
+        &'b mut self,
+        embedder: &'a str,
+        semantic_ratio: f32,
+    ) -> &'b mut SearchQuery<'a> {
+        self.hybrid = Some(HybridSearch {
+            embedder,
+            semantic_ratio,
+        });
+        self
+    }
+
     #[must_use]
     pub fn build(&mut self) -> SearchQuery<'a> {
         self.clone()
@@ -582,6 +619,7 @@ mod tests {
     use meilisearch_test_macro::meilisearch_test;
     use serde::{Deserialize, Serialize};
     use serde_json::{json, Map, Value};
+    use std::time::Duration;
 
     #[derive(Debug, Serialize, Deserialize, PartialEq)]
     struct Nested {
@@ -623,9 +661,15 @@ mod tests {
             .await?;
         let t2 = index.set_sortable_attributes(["title"]).await?;
 
-        t2.wait_for_completion(client, None, None).await?;
-        t1.wait_for_completion(client, None, None).await?;
-        t0.wait_for_completion(client, None, None).await?;
+        // the vector search has longer indexing times leading to the timeout being triggered
+        let timeout = if cfg!(feature = "experimental-vector-search") {
+            Some(Duration::from_secs(120))
+        } else {
+            None
+        };
+        t2.wait_for_completion(client, None, timeout).await?;
+        t1.wait_for_completion(client, None, timeout).await?;
+        t0.wait_for_completion(client, None, timeout).await?;
 
         Ok(())
     }
@@ -1172,6 +1216,89 @@ mod tests {
 
             assert!(!result.hits.is_empty());
         }
+
+        Ok(())
+    }
+
+    #[cfg(feature = "experimental-vector-search")]
+    #[meilisearch_test]
+    async fn test_hybrid(client: Client, index: Index) -> Result<(), Error> {
+        log::warn!("You are executing the vector search test. This WILL take a while and might lead to timeouts in other tests. You can disable this testcase by not enabling the `experimental-vector-search`-feature and running this ");
+        // enable vector searching and configure an embedder
+        let features = crate::ExperimentalFeatures::new(&client)
+            .set_vector_store(true)
+            .update()
+            .await
+            .expect("could not enable the vector store");
+        assert_eq!(features.vector_store, true);
+        let embedder_setting = crate::Embedder::HuggingFace(crate::HuggingFaceEmbedderSettings {
+            model: "BAAI/bge-base-en-v1.5".into(),
+            revision: None,
+            document_template: Some("{{ doc.value }}".into()),
+        });
+        let t3 = index
+            .set_settings(&crate::Settings {
+                embedders: Some(HashMap::from([("default".to_string(), embedder_setting)])),
+                ..crate::Settings::default()
+            })
+            .await?;
+        t3.wait_for_completion(&client, None, None).await?;
+
+        setup_test_index(&client, &index).await?;
+
+        // "zweite" = "second" in german
+        // => an embedding should be able to detect that this is equivalent, but not the regular search
+        let results: SearchResults<Document> = index
+            .search()
+            .with_query("Facebook")
+            .with_hybrid("default", 1.0) // entirely rely on semantic searching
+            .execute()
+            .await?;
+        assert_eq!(results.hits.len(), 1);
+        assert_eq!(
+            &Document {
+                id: 1,
+                value: S("dolor sit amet, consectetur adipiscing elit"),
+                kind: S("text"),
+                number: 10,
+                nested: Nested { child: S("second") },
+            },
+            &results.hits[0].result
+        );
+        let results: SearchResults<Document> = index
+            .search()
+            .with_query("zweite")
+            .with_hybrid("default", 0.0) // no semantic searching => no matches
+            .execute()
+            .await?;
+        assert_eq!(results.hits.len(), 0);
+
+        // word that has a typo => would have been found via traditional means
+        // if entirely relying on semantic searching, no result is found
+        let results: SearchResults<Document> = index
+            .search()
+            .with_query("lohrem")
+            .with_hybrid("default", 1.0)
+            .execute()
+            .await?;
+        assert_eq!(results.hits.len(), 0);
+        let results: SearchResults<Document> = index
+            .search()
+            .with_query("lohrem")
+            .with_hybrid("default", 0.0)
+            .execute()
+            .await?;
+        assert_eq!(results.hits.len(), 1);
+        assert_eq!(
+            &Document {
+                id: 0,
+                value: S("Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."),
+                kind: S("text"),
+                number: 0,
+                nested: Nested { child: S("first") }
+            },
+            &results.hits[0].result.id
+        );
 
         Ok(())
     }
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -77,6 +77,7 @@ pub struct OpenapiEmbedderSettings {
     /// The openapi model name
     /// Example: `text-embedding-ada-002`
     pub model: String,
+    pub dimensions: usize,
     /// if present, document_template must be a [Liquid template](https://shopify.github.io/liquid/).
     /// Use `{{ doc.attribute }}` to access document field values.
     /// Meilisearch also exposes a `{{ fields }}` array containing one object per document field, which you may access with `{{ field.name }}` and `{{ field.value }}`.
@@ -90,7 +91,7 @@ pub struct OpenapiEmbedderSettings {
 #[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq, Copy)]
 pub struct CustomEmbedderSettings {
     /// dimensions of your custom embedding
-    pub dimensions: u32,
+    pub dimensions: usize,
 }
 
 /// Struct reprensenting a set of settings.
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -55,6 +55,8 @@ pub struct HuggingFaceEmbedderSettings {
     /// the BERT embedding model you want to use from HuggingFace
     /// Example: `bge-base-en-v1.5`
     pub model: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub revision: Option<String>,
     /// if present, document_template must be a [Liquid template](https://shopify.github.io/liquid/).
     /// Use `{{ doc.attribute }}` to access document field values.
     /// Meilisearch also exposes a `{{ fields }}` array containing one object per document field, which you may access with `{{ field.name }}` and `{{ field.value }}`.
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -34,6 +34,63 @@ pub struct FacetingSettings {
     pub max_values_per_facet: usize,
 }
 
+#[cfg(feature = "experimental-vector-search")]
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase", tag = "source")]
+pub enum Embedder {
+    /// Compute embeddings locally.
+    /// This is a resource-intensive operation and might affect indexing performance.
+    HuggingFace(HuggingFaceEmbedderSettings),
+    /// Use OpenAi's API to generate embeddings
+    OpenAi(OpenapiEmbedderSettings),
+    /// Provide custom embeddings.
+    /// In this case, you must manually update your embeddings when adding, updating, and removing documents to your index.
+    UserProvided(CustomEmbedderSettings),
+}
+
+#[cfg(feature = "experimental-vector-search")]
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct HuggingFaceEmbedderSettings {
+    /// the BERT embedding model you want to use from HuggingFace
+    /// Example: `bge-base-en-v1.5`
+    pub model: String,
+    /// if present, document_template must be a [Liquid template](https://shopify.github.io/liquid/).
+    /// Use `{{ doc.attribute }}` to access document field values.
+    /// Meilisearch also exposes a `{{ fields }}` array containing one object per document field, which you may access with `{{ field.name }}` and `{{ field.value }}`.
+    ///
+    /// For best results, use short strings indicating the type of document in that index, only include highly relevant document fields, and truncate long fields.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub document_template: Option<String>,
+}
+
+#[cfg(feature = "experimental-vector-search")]
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct OpenapiEmbedderSettings {
+    /// API key used to authorize against OpenAI.
+    /// [Generate an API key](https://platform.openai.com/api-keys) from your OpenAI account.
+    /// Use [tier 2 keys](https://platform.openai.com/docs/guides/rate-limits/usage-tiers?context=tier-two) or above for optimal performance.
+    pub api_key: String,
+    /// The openapi model name
+    /// Example: `text-embedding-ada-002`
+    pub model: String,
+    /// if present, document_template must be a [Liquid template](https://shopify.github.io/liquid/).
+    /// Use `{{ doc.attribute }}` to access document field values.
+    /// Meilisearch also exposes a `{{ fields }}` array containing one object per document field, which you may access with `{{ field.name }}` and `{{ field.value }}`.
+    ///
+    /// For best results, use short strings indicating the type of document in that index, only include highly relevant document fields, and truncate long fields.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub document_template: Option<String>,
+}
+
+#[cfg(feature = "experimental-vector-search")]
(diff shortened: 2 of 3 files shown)


curquiza:
Hello @CommanderStorm 

Can you rebase your branch? We made changes recently to improve the library

Sorry for the inconvenience, I try to review your PR as soon as possible

CommanderStorm:
> Can you rebase your branch?

Done

src/search.rs
```diff
@@ -654,9 +693,15 @@ mod tests {
             .await?;
         let t2 = index.set_sortable_attributes(["title"]).await?;
 
-        t2.wait_for_completion(client, None, None).await?;
-        t1.wait_for_completion(client, None, None).await?;
-        t0.wait_for_completion(client, None, None).await?;
+        // the vector search has longer indexing times leading to the timeout being triggered
+        let timeout = if cfg!(feature = "experimental-vector-search") {
+            Some(Duration::from_secs(120))
```
irevoire, line 698: I did a few tests, and on my side it never took 120s, the quickest execution took 150s but most of the time I was over 200s

irevoire (changes requested):
Hey @CommanderStorm,

Code-wise, the PR is very nice and well-documented; I love it!

-----

But I’m hitting way too many timeouts to really help you; sorry 😖 

I don’t think we’ll be able to merge this PR with tests that take about 10 minutes. Could you mock Meilisearch as we did here:
https://github.com/meilisearch/meilisearch-rust/blob/cb325340ea60a10030f304f78aeae4942eeb3d33/src/tasks.rs#L902-L924

That way, we simply ensure that meilisearch-rust sends a valid payload and hope meilisearch works as expected.

> userProvided seens more brittle, but we may want change to this instead

Or we could do that and actually send the payload to meilisearch.
Or do both; let me know what you prefer, but I would very much like the tests to not take that much time to run 😖 

> Where in the Meilisearch codebase could I find an e2e-test how to use feature?

I don’t think there is. I believe you’re right, and we only wrote tests for user-provided vectors

> introduces the experimental-vector-search-feature flag

I don’t know if this is required, @curquiza. Could we simply expose the feature as-is instead of hiding it behind a feature flag (that will make it harder to test and use).
Since it doesn’t add any dependency, I don’t see much point in putting it behind a feature flag

curquiza:
> I don’t know if this is required, @curquiza. Could we simply expose the feature as-is instead of hiding it behind a feature flag (that will make it harder to test and use).

Yes, that's ok not to do feature flag 😊

CommanderStorm:
> I did a few tests, and on my side it never took 120s, the quickest execution took 150s but most of the time I was over 200s

Time if migrated to `userProvided` seems fine in CI (this is likely the slowest machine running the tests).

> Or we could do that and actually send the payload to meilisearch.

I think keeping the testcases from requiring a active internet connection is better as otherwise the test might be unnessesarily flaky in CI.

> Could you mock Meilisearch as we did here

I can add a testcase where I mock the routes.
I am unsure if you actually would want to mock this for an experimental feature (= where the api might change => requiring changes)

I have tweaked a bunch with the vectors available and can't get the `test_hybrid` testcase to work without fully using the same dataset as in [`tests/search/hybrid.rs`](https://github.com/meilisearch/meilisearch/blob/938149f8144edd31365105000c6279634b9437ef/meilisearch/tests/search/hybrid.rs).
It seems to always return everything when I set `semantic_ratio=1.0`... Is this operating as intended?
I have tried similar 2D vectors as the [upstream test](https://github.com/meilisearch/meilisearch/blob/main/meilisearch/tests/search/hybrid.rs) and went as far as to use 1D vectors with considerable (1/10/1k/1m) spread (=>something that as far as I understand embeddings should not match)
=> I am missing something. Could you point me into the right direction? 😅

(I can obviously steal the testcase from [`tests/search/hybrid.rs`](https://github.com/meilisearch/meilisearch/blob/938149f8144edd31365105000c6279634b9437ef/meilisearch/tests/search/hybrid.rs), but for longer-term maintainability I would like this testcase to not be such an oddball ^^)

CommanderStorm:
Thanks for the pinging (github does not give me notifications for this) ^^
No need to worry. If this takes a few months that is fine.

src/search.rs
```diff
@@ -753,7 +802,8 @@ mod tests {
                 value: S("dolor sit amet, consectetur adipiscing elit"),
                 kind: S("text"),
                 number: 10,
-                nested: Nested { child: S("second") }
+                nested: Nested { child: S("second") },
+                _vectors: HashMap::from([(S("default"), vec![2000.0])]),
```
CommanderStorm, line 806: Note to self/TODO:

`retrieveVectors` needs to be configurable and set to be able to get reuse the type
Docs: https://github.com/meilisearch/meilisearch/releases/tag/v1.9.0

curquiza (changes requested):
@CommanderStorm thank you again! Can you add the `rest` and `ollama` models we added in [v1.8.0](https://github.com/meilisearch/meilisearch/releases/v1.8.0)?
Sorry for the late notice again!
And thank you for your involvement 🙏

src/search.rs
curquiza: Do you want to do it here, or in another PR?

CommanderStorm: Will do that here (would not make sense in another PR imo).
I am going to do that towards the end of the week

irevoire:
Hey @CommanderStorm, since we introduced a new `rest` embedder, I think we could write better tests by mocking the rest embedder and ensuring it works with meilisearch; here's some example I wrote earlier this week in meilisearch: https://github.com/meilisearch/meilisearch/pull/4755/commits/2141cb3b69f52ca6fcf18d0b89df1d8601a22f85

src/search.rs
```diff
@@ -1274,4 +1366,110 @@ mod tests {
...
+            .with_query("lorem ipsum")
+            .with_retrieve_vectors(false)
+            .execute().await?;
+        assert_eq!(results.hits.len(), 1);
+        assert_eq!(results.hits[0].result._vectors, None);
+        Ok(())
+    }
+
+    #[meilisearch_test]
+    async fn test_hybrid(client: Client, index: Index) -> Result<(), Error> {
```
CommanderStorm, line 1413: Small status update:
- [x] `with_vectors`
- [x] Ollama
- [x] Rest
- [ ] refactoring `test_hybrid` to mock the network calls made instead of failing to figgure out how the local embedder really works

src/search.rs
```diff
@@ -776,7 +857,10 @@ mod tests {
                 value: S("dolor sit amet, consectetur adipiscing elit"),
                 kind: S("text"),
                 number: 10,
-                nested: Nested { child: S("second") }
+                nested: Nested { child: S("second") },
+                // TODO: This is likely a bug upstream. vectors should not be present
+                // => correct would be `_vectors: None`
+                _vectors: Some(Vectors::from(&[2000.0])),
```
CommanderStorm, line 863: TODO(Frank): Report this upstream

irevoire:
```suggestion
                nested: Nested { child: S("second") },
                _vectors: Some(Vectors::from(&[2000.0])),
```

Hey, this is not a bug; you can remove the comment

src/search.rs
```diff
@@ -666,6 +730,29 @@ mod tests {
         kind: String,
         number: i32,
         nested: Nested,
+        #[serde(default)]
+        _vectors: Option<Vectors>,
+    }
+
+    #[derive(Debug, Serialize, Deserialize, PartialEq)]
+    struct Vector {
+        embeddings: Vec<Vec<f32>>,
```
irevoire, line 739: Here meilisearch can return a `Vec<f32>` directly.

You could do something like that;
```rust
    #[derive(Debug, Serialize, Deserialize, PartialEq)]
    struct Vector {
        embeddings: SingleOrMultipleVectors,
        regenerate: bool,
    }

    #[derive(Serialize, Deserialize)]
    #[serde(untagged)]
    enum SingleOrMultileVectors {
        Single(Vec<f32>),
        Multiple(Vec<Vec<f32>>),
    }
```

irevoire (changes requested):
Hey @CommanderStorm,

The PR seems really good; I think we're almost ready to merge.

I think you missed one issue in your test; your definition of what a vector is was wrong, and thus, serde was ignoring most vectors when deserializing the documents.
I left a comment on the subject.

To write the test you wanted to write initially, where the vectors appear as `None,` you should enable the experimental feature first.

This brings me to another point; I think every test using the feature should start by enabling it just in case they run first.

**AND** we should absolutely remove these two tests:
https://github.com/meilisearch/meilisearch-rust/blob/c673c8d3889695b954acc7db09fc568301630097/src/features.rs#L108-L132

That will mess with everything

CommanderStorm:
I think I have adressed the requested changes ^^
Sorry for the long delay, I was on vacation and then a bit buisy with other work. Hope you understand.

> AND we should absolutely remove these two tests

I have only removed the test which disables the vector store, but have kept the test which `.set_vector_store(true)`'s.
I have also improved a test a bit to also set, but this should be minor ^^

Next time, I will split such PRs earlier ^^

src/settings.rs
```diff
@@ -36,7 +36,260 @@ pub struct FacetingSettings {
...
+///   path_to_embeddings: vec!["data".to_string()],
+///   embedding_object: vec!["embedding".to_string()],
+/// };
+/// # let expected = r#"{"url":"http://localhost:12345/api/v1/embed","apiKey":"SOURCE_API_KEY","dimensions":512,"documentTemplate":"A document titled {{doc.title}} whose description starts with {{doc.overview|truncatewords: 20}}","inputField":["data","text"],"inputType":"text","query":{"dimensions":512,"model":"MODEL_NAME"},"pathToEmbeddings":["data"],"embeddingObject":["embedding"]}"#;
+/// # let expected: GenericRestEmbedderSettings = serde_json::from_str(expected).unwrap();
+/// # assert_eq!(embedder_setting, expected);
+/// ```
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct GenericRestEmbedderSettings {
```
CommanderStorm, line 217: Needs to be adapted to account for the breaking changes in `v1.10.0`:

https://github.com/meilisearch/meilisearch/releases/tag/v1.10.0#:~:text=Changing%20the%20parameters%20of%20the%20REST%20API

CommanderStorm: https://github.com/meilisearch/meilisearch-rust/pull/554/commits/73981850e687a87d7d317dbef7b5ad969e77f78b

src/settings.rs
```diff
@@ -36,7 +36,265 @@ pub struct FacetingSettings {
...
+/// ```
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct OpenapiEmbedderSettings {
+    /// API key used to authorize against OpenAI.
+    /// [Generate an API key](https://platform.openai.com/api-keys) from your OpenAI account.
+    /// Use [tier 2 keys](https://platform.openai.com/docs/guides/rate-limits/usage-tiers?context=tier-two) or above for optimal performance.
+    pub api_key: String,
+    /// The openapi model name
+    /// Default: `text-embedding-ada-002`
```
CommanderStorm, line 124:
```suggestion
    /// Default: `text-embedding-3-small`
```

src/settings.rs
```diff
@@ -36,7 +36,265 @@ pub struct FacetingSettings {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub dimensions: Option<usize>,
+    /// Use it to customize the data you send to the embedder. It is highly recommended you configure a custom template for your documents.
+    ///
+    /// if present, `document_template` must be a [Liquid template](https://shopify.github.io/liquid/).
+    /// Use `{{ doc.attribute }}` to access document field values.
+    /// Meilisearch also exposes a `{{ fields }}` array containing one object per document field, which you may access with `{{ field.name }}` and `{{ field.value }}`.
+    ///
+    /// For best results, use short strings indicating the type of document in that index, only include highly relevant document fields, and truncate long fields.
+    /// Example: `"A document titled '{{doc.title}}' whose description starts with {{doc.overview|truncatewords: 20}}"`
```
CommanderStorm, line 251:
```suggestion
    /// Example: `"A document titled '{{doc.title}}' whose description starts with {{doc.overview|truncatewords: 20}}"`
    /// 
    /// Default:
    /// ```raw
    /// {% for field in fields %}
    /// {% if field.is_searchable and not field.value == nil %}
    /// {{ field.name }}: {{ field.value }}\n
    /// {% endif %}
    /// {% endfor %}
    /// ```
```

src/settings.rs
```diff
@@ -36,7 +36,265 @@ pub struct FacetingSettings {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub revision: Option<String>,
+    /// Use it to customize the data you send to the embedder. It is highly recommended you configure a custom template for your documents.
+    ///
+    /// if present, `document_template` must be a [Liquid template](https://shopify.github.io/liquid/).
+    /// Use `{{ doc.attribute }}` to access document field values.
+    /// Meilisearch also exposes a `{{ fields }}` array containing one object per document field, which you may access with `{{ field.name }}` and `{{ field.value }}`.
+    ///
+    /// For best results, use short strings indicating the type of document in that index, only include highly relevant document fields, and truncate long fields.
+    /// Example: "A document titled '{{doc.title}}' whose description starts with {{doc.overview|truncatewords: 20}}"
```
CommanderStorm, line 95:
```suggestion
    /// Example: `"A document titled '{{doc.title}}' whose description starts with {{doc.overview|truncatewords: 20}}"`
    /// 
    /// Default:
    /// ```raw
    /// {% for field in fields %}
    /// {% if field.is_searchable and not field.value == nil %}
    /// {{ field.name }}: {{ field.value }}\n
    /// {% endif %}
    /// {% endfor %}
    /// ```
```

src/settings.rs
```diff
@@ -36,7 +36,265 @@ pub struct FacetingSettings {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub dimensions: Option<usize>,
+    /// Use it to customize the data you send to the embedder. It is highly recommended you configure a custom template for your documents.
+    ///
+    /// if present, `document_template` must be a [Liquid template](https://shopify.github.io/liquid/).
+    /// Use `{{ doc.attribute }}` to access document field values.
+    /// Meilisearch also exposes a `{{ fields }}` array containing one object per document field, which you may access with `{{ field.name }}` and `{{ field.value }}`.
+    ///
+    /// For best results, use short strings indicating the type of document in that index, only include highly relevant document fields, and truncate long fields.
+    /// Example: "A document titled '{{doc.title}}' whose description starts with {{doc.overview|truncatewords: 20}}"
```
CommanderStorm, line 137:
```suggestion
    /// Example: `"A document titled '{{doc.title}}' whose description starts with {{doc.overview|truncatewords: 20}}"`
    /// 
    /// Default:
    /// ```raw
    /// {% for field in fields %}
    /// {% if field.is_searchable and not field.value == nil %}
    /// {{ field.name }}: {{ field.value }}\n
    /// {% endif %}
    /// {% endfor %}
    /// ```
```

src/settings.rs
```diff
@@ -36,7 +36,265 @@ pub struct FacetingSettings {
...
+    /// | `snowflake-arctic-embed` | varies       | [View model](https://ollama.com/library/snowflake-arctic-embed) |
+    pub model: String,
+    /// Use it to customize the data you send to the embedder. It is highly recommended you configure a custom template for your documents.
+    ///
+    /// if present, `document_template` must be a [Liquid template](https://shopify.github.io/liquid/).
+    /// Use `{{ doc.attribute }}` to access document field values.
+    /// Meilisearch also exposes a `{{ fields }}` array containing one object per document field, which you may access with `{{ field.name }}` and `{{ field.value }}`.
+    ///
+    /// For best results, use short strings indicating the type of document in that index, only include highly relevant document fields, and truncate long fields.
+    /// Example: "A document titled '{{doc.title}}' whose description starts with {{doc.overview|truncatewords: 20}}"
```
CommanderStorm, line 188:
```suggestion
    /// Example: `"A document titled '{{doc.title}}' whose description starts with {{doc.overview|truncatewords: 20}}"`
    /// 
    /// Default:
    /// ```raw
    /// {% for field in fields %}
    /// {% if field.is_searchable and not field.value == nil %}
    /// {{ field.name }}: {{ field.value }}\n
    /// {% endif %}
    /// {% endfor %}
    /// ```
```

CommanderStorm (commented):
This updates the PR to the breaking `1.11` changes.

CommanderStorm:
[`binaryQuantized`](https://www.meilisearch.com/docs/reference/api/settings#binaryquantized) will have to be in a different PR, as according to the docs this is only avaliable for updating via `PATCH`, which is currently not supported.

CommanderStorm:
@irevoire sorry to ping you.
I think it would be much more maintainable to do this work in different PRs.

Is merging this feature in the experimental stage still something you are interested in, or would this feature need to be stabilised first?

src/search.rs
```diff
@@ -350,6 +374,21 @@ pub struct SearchQuery<'a, Http: HttpClient> {
...
+    /// Defines what vectors an userprovided embedder has gotten for semantic searching
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub vector: Option<&'a [f32]>,
+
+    /// Defines whether vectors for semantic searching are returned in the search results
+    /// Can Significantly increase the response size.
+    ///
+    /// **Default: `false`**
+    #[serde(skip_serializing_if = "Option::is_none")]
+    retrieve_vectors: Option<bool>,
```
CommanderStorm, line 391:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub retrieve_vectors: Option<bool>,
```

src/search.rs
```diff
@@ -785,7 +883,8 @@ mod tests {
                 value: S("dolor sit amet, consectetur adipiscing elit"),
                 kind: S("text"),
                 number: 10,
-                nested: Nested { child: S("second") }
+                nested: Nested { child: S("second") },
+                _vectors: Some(Vectors::from(&[2000.0])),
```
CommanderStorm, line 887: FYI: the integration testcases are failing on the `:latest` meilisearch image as the following experimental feature is now removed

- https://github.com/meilisearch/meilisearch-rust/pull/638

CommanderStorm:
```suggestion
                _vectors: None,
```

src/search.rs
```diff
@@ -1307,4 +1414,77 @@ mod tests {
...
+        use crate::settings::{Embedder, UserProvidedEmbedderSettings};
+        let embedder_setting =
+            Embedder::UserProvided(UserProvidedEmbedderSettings { dimensions: 1 });
+        let t3 = index
+            .set_settings(&crate::settings::Settings {
+                embedders: Some(HashMap::from([("default".to_string(), embedder_setting)])),
+                ..crate::settings::Settings::default()
+            })
+            .await?;
+        t3.wait_for_completion(&client, None, None).await?;
```
CommanderStorm, line 1429:
```suggestion
        index
            .set_settings(&crate::settings::Settings {
                embedders: Some(HashMap::from([("default".to_string(), embedder_setting)])),
                ..crate::settings::Settings::default()
            })
            .await?
            .wait_for_completion(&client, None, None)
            .await?;
```

Akagi201:
Any update to this PR? It's very useful for AI search

CommanderStorm:
From a features/tests/documentation POV nothing is blocking merging.
I suspect that meilisearch noticed that the other integrations are much more used and is allocating reviewer capacity accordingly.

In terms of docs, my last sync with their docs/features was two months ago. If there is drift, feel free to provide a PR to this PR or copy my changes into a new PR.
At this point, I have abandoned the idea of AI search due the increased latency (I don't run on fast hardware) being a noticeably worse experience for my use-case.
Since said functionality is stable now, I am leaving this open. If the team wants some change, I think I can do that too, but only if there is a hope of this leading to merging, which at the moment does not look promising.

You can use this via overriding what meilisearch means
```diff
- meilisearch-sdk = "1"
+ meilisearch-sdk = { git = "https://github.com/commanderstorm/meilisearch-rust.git", branch = "vector-search-embedder" }

**Comment by awyl** [comment]
Hi,

I ran into this error when I try to create an index with settings provided in this PR.

Meilisearch: 1.14

```rs
Settings::new()
    .with_embedders(HashMap::from([(
        "default",
        Embedder::OpenAI(OpenAIEmbedderSettings {
            api_key: "...",
            model: Some("text-embedding-3-small".to_string()),
            dimensions: Some(1536),
            document_template_max_bytes: Some(6000),
            ..Default::default()
        }),
    )]))
```

Error:
```
Meilisearch invalid_request: unknown: Unknown value `openAI` at `.embedders.default.source`: did you mean 
`openAi`? expected one of `openAi`, `huggingFace`, `ollama`, `userProvided`, `rest`, `composite`. https://docs.me
ilisearch.com/errors#invalid_settings_embedders
```

src/settings.rs
```diff
@@ -36,14 +36,355 @@ pub struct FacetingSettings {
...
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase", tag = "source")]
+pub enum Embedder {
+    /// Compute embeddings inside meilisearch with models from [HuggingFace](https://huggingface.co/).
+    /// You may be able to significantly improve performance by [compiling a CUDA-compatible Meilisearch binary](https://www.meilisearch.com/docs/guides/ai/computing_hugging_face_embeddings_gpu).
+    /// This is a resource-intensive operation and might affect indexing performance negatively.
+    HuggingFace(HuggingFaceEmbedderSettings),
+    /// Use OpenAI's API to generate embeddings
+    /// Depending on hardware, this is a
+    OpenAI(OpenAIEmbedderSettings),
```
CommanderStorm, line 49: fixes above bug report

```suggestion
    OpenAi(OpenAIEmbedderSettings),
```

src/settings.rs
```diff
@@ -36,14 +36,355 @@ pub struct FacetingSettings {
...
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase", tag = "source")]
+pub enum Embedder {
+    /// Compute embeddings inside meilisearch with models from [HuggingFace](https://huggingface.co/).
+    /// You may be able to significantly improve performance by [compiling a CUDA-compatible Meilisearch binary](https://www.meilisearch.com/docs/guides/ai/computing_hugging_face_embeddings_gpu).
+    /// This is a resource-intensive operation and might affect indexing performance negatively.
+    HuggingFace(HuggingFaceEmbedderSettings),
+    /// Use OpenAI's API to generate embeddings
+    /// Depending on hardware, this is a
+    OpenAi(OpenAIEmbedderSettings),
```
CommanderStorm, line 49:
```suggestion
    /// Compute embeddings inside meilisearch with models from [HuggingFace](https://huggingface.co/).
    ///
    /// You may be able to significantly improve performance by [compiling a CUDA-compatible Meilisearch binary](https://www.meilisearch.com/docs/guides/ai/computing_hugging_face_embeddings_gpu).
    /// This is a resource-intensive operation and might affect indexing performance negatively.
    HuggingFace(HuggingFaceEmbedderSettings),
    /// Use OpenAI's API to generate embeddings
    ///
    /// Depending on your hardware, this network request may be faster
    OpenAi(OpenAIEmbedderSettings),
```

src/settings.rs
```diff
@@ -36,14 +36,357 @@ pub struct FacetingSettings {
...
+/// # let expected: GenericRestEmbedderSettings = serde_json::from_value(expected).unwrap();
+/// # assert_eq!(embedder_setting, expected);
+/// ```
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct GenericRestEmbedderSettings {
+    /// Mandatory, full URL to the embedding endpoint
+    ///
+    /// Must be parseable as a URL.
+    /// If not specified, [Meilisearch](https://www.meilisearch.com/) (**not the sdk you are currently using**) will try to fetch the `MEILI_OLLAMA_URL` environment variable
```
CommanderStorm, line 287:
```suggestion
```

CommanderStorm:
@awyl could you retry please? Should be fixed now

awyl:
> @awyl could you retry please? Should be fixed now

Thank you, it works perfectly.

CommanderStorm:
The status I have written above is still accurate

> From a features/tests/documentation POV nothing is blocking merging. I suspect that meilisearch noticed that the other integrations are much more used and is allocating reviewer capacity accordingly.
> 
> In terms of docs, my last sync with their docs/features was two months ago. If there is drift, feel free to provide a PR to this PR or copy my changes into a new PR. At this point, I have abandoned the idea of AI search due the increased latency (I don't run on fast hardware) being a noticeably worse experience for my use-case.
> 
> Since said functionality is stable now, I am leaving this open. If the team wants some change, I think I can do that too, but only if there is a hope of this leading to merging, which at the moment does not look promising.
> 
> You can use this via overriding what meilisearch means
> 
> ```diff
> - meilisearch-sdk = "1"
> + meilisearch-sdk = { git = "https://github.com/commanderstorm/meilisearch-rust.git", branch = "vector-search-embedder" }
> ```

PR #562, opened 2024-04-08
Fix proximity precision codesample

ellnix:
Noticed this while browsing documentation.


irevoire (approved):
Nice, thank you

bors merge

Issue #530, opened 2023-10-12
Library seems unusable on MacOS 14.0


Jasperav:
There is an annoying bug with MacOS 14.0 which causes binaries to fail: https://github.com/rust-lang/cargo/issues/12670

This is a PR which fixes the bug: https://github.com/alexcrichton/curl-rust/pull/530. Now it seems easy: update to the latest CURL and it should work. In my project, I got this output running `cargo tree`:

```
│   ├── meilisearch-sdk v0.24.2
│   │   ├── async-trait v0.1.73 (proc-macro) (*)
│   │   ├── either v1.9.0 (*)
│   │   ├── futures v0.3.28 (*)
│   │   ├── futures-io v0.3.28
│   │   ├── isahc v1.7.2
│   │   │   ├── async-channel v1.9.0
│   │   │   │   ├── concurrent-queue v2.2.0
│   │   │   │   │   └── crossbeam-utils v0.8.16 (*)
│   │   │   │   ├── event-listener v2.5.3
│   │   │   │   └── futures-core v0.3.28
│   │   │   ├── castaway v0.1.2
│   │   │   ├── crossbeam-utils v0.8.16 (*)
│   │   │   ├── curl v0.4.44
│   │   │   │   ├── curl-sys v0.4.65+curl-8.2.1
```

So `isahc` has a dependency on an old version of curl. It looks like a project which isn't maintained anymore. So unless I am totally wrong, this library can not be used on MacOS 14.0 anymore if none of those two things happen:
- Fork the `isahc` library and update `curl` there and use it here
- Switch to a maintained library which uses up to date `curl` version

c1wren:
This issue would be resolved if https://github.com/meilisearch/meilisearch-rust/pull/524 got merged, which doesn't use CURL at all. I've gotten around the issue by referencing that git branch in the pull request instead of crates.io version in my project.

PR #563, opened 2024-04-11
Use reqwest by default

irevoire:
# Pull Request

## Related issue
Fixes https://github.com/meilisearch/meilisearch-rust/issues/530 because we won’t depend on curl anymore

## What does this PR do?
- Get rid of `isahc` by default in favor of `reqwest`, which is way more used in the ecosystem

## PR checklist
- [x] Create an example changing the default HTTP client
- [x] Simplify the `HttpClient` trait to only require one method to be implemented
- [x] Stores a basic `reqwest::Client` in the `meilisearch_sdk::Client` instead of re-creating it from scratch for every request
- [x] Double check it works with wasm
- [x] Remove all the unwraps in `request`
- [x] Do not use the `User-Agent` when in wasm as it is often blocked by the browser


gibbz00:
It is as if https://github.com/meilisearch/meilisearch-rust/pull/524 doesn't exist

irevoire:
I did look at it, but I don't have much time to work on this. There are 4 PRs doing pretty much the same thing, and I found your trait to be more complex than the one I initially wanted sorry

irevoire (commented):
bors merge

curquiza (approved):

curquiza:
bors merge

PR #576, opened 2024-04-21
with_distrinct_attribute accepts None

### LINKED ISSUE #368: with_distrinct_attribute should accept `None` as a value.
**Description**
Currently `with_distinct_attribute` does not accept `None` as a value. Meaning that once you have set one distinct attribute you need to either user `reset_distinct_attribute` or create an empty string as value.

https://github.com/meilisearch/meilisearch-rust/blob/6c3a1d2c4acd482c4d57837bb4ee1f74c1d2c0b3/src/settings.rs#L188-L193

**Expected behavior**
`None` should be a valid type

**Current behavior**
It is not

NoodleSamaChan:
# Pull Request

## Related issue
Fixes #368

## What does this PR do?
- changed with_distinct_attribute so that it accepts None

## PR checklist
- [X] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?
- [X] Have you read the contributing guidelines?
- [X] Have you made sure that the title is accurate and descriptive of the changes?

### DIFF
--- a/src/settings.rs
+++ b/src/settings.rs
@@ -217,19 +217,18 @@ impl Settings {
     }
 
     #[must_use]
-    pub fn with_distinct_attribute(self, distinct_attribute:  Option< impl AsRef<str>>) -> Settings {
+    pub fn with_distinct_attribute(self, distinct_attribute: Option<impl AsRef<str>>) -> Settings {
         if let Some(distinct_attribute) = distinct_attribute {
-                Settings {
+            Settings {
                 distinct_attribute: Some(Some(distinct_attribute.as_ref().to_string())),
                 ..self
-                }
-            } else {
-                Settings {
+            }
+        } else {
+            Settings {
                 distinct_attribute: Some(None),
                 ..self
-                }
-
-            } 
+            }
+        }
     }
 
     #[must_use]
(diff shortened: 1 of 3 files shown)


src/settings.rs
```diff
@@ -217,10 +217,17 @@ impl Settings {
...
+        if let Some(distinct_attribute) = distinct_attribute {
+            Settings {
+                distinct_attribute: Some(Some(distinct_attribute.as_ref().to_string())),
+                ..self
+            }
+        } else {
+            Settings {
+                distinct_attribute: Some(None),
+                ..self
+            }
```
irevoire, line 230:
```suggestion
    pub fn with_distinct_attribute(self, distinct_attribute: Option<impl AsRef<str>>) -> Settings {
        Settings {
            distinct_attribute: Some(distinct_attribute.map(|distinct| distinct.to_string())),
            ..self
        }

```

Use the [`map`](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.map) method on options

irevoire (changes requested):

irevoire (approved):
Thanks,

bors merge

Issue #575, opened 2024-04-18
Unwrap causes panic instead of handling the Result


kingsleyh:
When checking if an index exists this unwrap returns None and causes a panic.

Instead could handle this instead of using unwrap() so that it always returns the Result 

file: meilisearch-sdk-0.25.0/src/client.rs:49:55
  
```rust
 fn parse_indexes_results_from_value(&self, value: &Value) -> Result<IndexesResults, Error> {
        let raw_indexes = value["results"].as_array().unwrap();

        let indexes_results = IndexesResults {
            limit: value["limit"].as_u64().unwrap() as u32,
            offset: value["offset"].as_u64().unwrap() as u32,
            total: value["total"].as_u64().unwrap() as u32,
            results: raw_indexes
                .iter()
                .map(|raw_index| Index::from_value(raw_index.clone(), self.clone()))
                .collect::<Result<_, _>>()?,
        };

        Ok(indexes_results)
    }
```

kingsleyh:
I was using an old version of meilisearch - once I upgraded to the latest this went away

irevoire:
Hey @kingsleyh 

Thanks for your issue; you're right that this is not a bug if you use the correct version of the SDK with the right version of Meilisearch.
But you're not wrong either; it shouldn't panic and return an error instead.

I'll reopen it but with the maintenance label instead of the bug one, thanks for the opening the issue!

PR #603, opened 2024-07-07
Stops panicking when getting indexes

postmeback:
Sorry for the wrong heading. It actually enhances the request mentioned on [575](https://github.com/meilisearch/meilisearch-rust/issues/575). 

Fixes: #575 

Requesting the maintainers to have a look at it.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -79,18 +79,34 @@ impl<Http: HttpClient> Client<Http> {
         &self,
         value: &Value,
     ) -> Result<IndexesResults<Http>, Error> {
-        let raw_indexes = value["results"].as_array().unwrap();
-
+        let raw_indexes = value["results"]
+            .as_array()
+            .ok_or_else(|| Error::ParseStringError("Missing or invalid 'results' field".to_string()))?;
+    
+        let limit = value["limit"]
+            .as_u64()
+            .ok_or_else(|| Error::ParseStringError("Missing or invalid 'limit' field".to_string()))? as u32;
+    
+        let offset = value["offset"]
+            .as_u64()
+            .ok_or_else(|| Error::ParseStringError("Missing or invalid 'offset' field".to_string()))? as u32;
+    
+        let total = value["total"]
+            .as_u64()
+            .ok_or_else(|| Error::ParseStringError("Missing or invalid 'total' field".to_string()))? as u32;
+    
+        let results = raw_indexes
+            .iter()
+            .map(|raw_index| Index::from_value(raw_index.clone(), self.clone()))
+            .collect::<Result<_, _>>()?;
+    
         let indexes_results = IndexesResults {
-            limit: value["limit"].as_u64().unwrap() as u32,
-            offset: value["offset"].as_u64().unwrap() as u32,
-            total: value["total"].as_u64().unwrap() as u32,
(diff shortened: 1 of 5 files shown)


src/errors.rs
```diff
@@ -17,6 +17,11 @@ pub enum Error {
     /// The Meilisearch server returned an invalid JSON for a request.
     #[error("Error parsing response JSON: {}", .0)]
     ParseError(#[from] serde_json::Error),
+
+    /// An error occurred while parsing the fields of the response JSON.
+    #[error("Error parsing fields: {0}")]
+    ParseStringError(String),
+
```
irevoire, line 24: Hey, I think it would be better if we didn't introduce any new error and instead returned hand-crafter `serde_json::Error` directly so our end user doesn't have to update their code to match this new error when it's, in fact, a deserialization error.

I’m ok with everything being a `custom_error`.
But if you really want, you can try to return more specific errors like`missing_field` and `invalid_value`

irevoire (changes requested):

src/errors.rs
postmeback: But, if everything becomes a `custom error`, 

the code change will be huge.

postmeback: I have made the changes, requesting for a review

src/errors.rs
```diff
@@ -15,8 +15,9 @@ pub enum Error {
     #[error(transparent)]
     MeilisearchCommunication(#[from] MeilisearchCommunicationError),
     /// The Meilisearch server returned an invalid JSON for a request.
-    #[error("Error parsing response JSON: {}", .0)]
-    ParseError(#[from] serde_json::Error),
+    #[error(transparent)]
+    SerdeParseError(#[from] serde_json::Error),
```
irevoire, line 19:
```suggestion
    #[error("Error parsing response JSON: {}", .0)]
    ParseError(#[from] serde_json::Error),
```

Hey, I don't think we should rename this error name, it's a breaking change that doesn't brings much to the end user

src/client.rs
```diff
@@ -79,16 +80,36 @@ impl<Http: HttpClient> Client<Http> {
...
+
+        let total = value["total"]
+            .as_u64()
+            .ok_or_else(|| serde_json::Error::custom("Missing or invalid 'total' field"))
+            .map_err(Error::SerdeParseError)? as u32;
+
+        let results = raw_indexes
+            .iter()
+            .map(|raw_index| Index::from_value(raw_index.clone(), self.clone()))
+            .collect::<Result<_, _>>()?;
```
irevoire, line 106: Perfect 👌

irevoire (changes requested):

src/client.rs
postmeback: Thanks 😃

postmeback:
Can you tell me how i can run the tests locally?

src/errors.rs
irevoire: Awesome thanks, sorry for the delay I was on holidays

irevoire (approved):
Hey @postmeback, awesome thanks a lot for your contribution

irevoire:
bors merge

postmeback:
Yaay ! 🪁 🪁 🪁

PR #616, opened 2024-09-01
Improve `IndexConfig` derive macro

funlennysub:
# Pull Request

## Related issue
None

## What does this PR do?
- A way to override the index name
- A way to change pagination settings
- Minor code improvements

## PR checklist


funlennysub:
Perhaps if there's no fields with `displayed` flag, it should default to `"*"` as meilisearch does, because from my understanding right now it will just set it to an empty array?

PR #666, opened 2025-05-02
Fix deleted_tasks field type in TaskDeletion struct

### LINKED ISSUE #665: TaskDeletion struct has incorrect type for field deleted_tasks
**Description**

When trying to call perform `get_task()` for a deletion task, I get an error: 
> Error parsing response JSON: invalid type: null, expected usize.

After getting the response object by calling the api manually and comparing it to the structs, I found out that `deletedTasks` was `null` in my response object, while the `deleted_tasks` field in struct `TaskDeletion` is `usize` and not `Option<usize>`
https://github.com/meilisearch/meilisearch-rust/blob/90a153c88112cddefc1936d3bef4763907211c2b/src/tasks.rs#L115-L121

According to the [API documentation](https://www.meilisearch.com/docs/reference/api/tasks#taskdeletion), the deleted_tasks can be null when the task is `enqueued` or `processing`.

coinmoles:
# Pull Request

## Related issue
Fixes #665

## What does this PR do?
- Change the type of `deleted_tasks` field in `TaskDeletion` from `usize` to `Option<usize>`
  - This should be `Option<usize>` since it can be null according to the [API documentation](https://www.meilisearch.com/docs/reference/api/tasks#taskdeletion)

## PR checklist

### DIFF
--- a/src/tasks.rs
+++ b/src/tasks.rs
@@ -116,7 +116,7 @@ pub struct TaskCancelation {
 #[serde(rename_all = "camelCase")]
 pub struct TaskDeletion {
     pub matched_tasks: usize,
-    pub deleted_tasks: usize,
+    pub deleted_tasks: Option<usize>,
     pub original_filter: String,
 }
 


Mubelotix (approved):
Nice catch! Thank you

PR #668, opened 2025-05-19
Serialize for MultiSearchResponse

milesgranger:
## What does this PR do?

This adds `serde::Serialize` to `MultiSearchResponse`.

I'm in a situation where I proxy requests between client and Meilisearch. During which I have authentication via OAuth and although client has logic in place, I want to inspect responses from Meilisearch and ensure the authenticated user is only getting data they are authorized for given group associations, whether indices or specific records in other index hits. 

The knee-jerk suggestion is to use different API keys. However, we have indices where for different users they have different permissions on what they can see and it's just not feasible/practical to duplicate indices/records into every combination of access rights.

## PR checklist

## Summary by CodeRabbit

- **New Features**
	- Improved support for exporting search results and related data in serialized formats.

### DIFF
--- a/src/search.rs
+++ b/src/search.rs
@@ -6,7 +6,7 @@ use serde::{de::DeserializeOwned, ser::SerializeStruct, Deserialize, Serialize, 
 use serde_json::{Map, Value};
 use std::collections::HashMap;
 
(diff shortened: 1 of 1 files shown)


Mubelotix (approved):
Alright, the drawbacks are minimal

PR #689, opened 2025-07-25
Use `tokio` sleep instead of blocking thread when using `reqwest` client

paolobarbolini:
# Pull Request

## Related issue

None

## What does this PR do?

This PR makes meilisearch-sdk use the much more efficient tokio sleep machinery when the `reqwest` HTTP client is used to make requests (we assume `reqwest` means that the user is using `tokio` because it only works with `tokio`) instead of spawing a new ad hoc blocking thread for each sleep request.

## PR checklist

## Summary by CodeRabbit

* **New Features**
  * Improved handling of asynchronous sleep operations, dynamically selecting the sleep mechanism based on the runtime environment.
  * Enhanced compatibility with different platforms, supporting both Tokio-based and non-Tokio environments.
* **Bug Fixes**
  * More reliable sleep behavior across various platforms and runtimes.
* **Tests**
  * Added and updated tests to ensure correct sleep functionality for each supported backend.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ futures-util = { version = "0.3.31", default-features = false, features = ["io"]
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 jsonwebtoken = { version = "9.3.1", default-features = false }
+tokio = { version = "1.38", optional = true, features = ["time"] }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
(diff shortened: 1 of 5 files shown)


Mubelotix:
Thanks for the quick response! I still do not see why anyone would want to use the old thread-blocking sleep implementation. Tokio's sleep should always be best, so we can remove the old one. Also, that makes SleepBackend redundant. I think we just need an internal `async fn sleep` that uses tokio's sleep except for wasm target. The tokio dependency can be enabled for everyone unless on wasm

PR #690, opened 2025-07-28
Make some fields in `*Stats` structs optional to support older servers

LukasKalbertodt:
These three fields were added in 1.13. If you want to support older servers, these fields need to be optional.

https://github.com/meilisearch/meilisearch/releases/tag/v1.13.0

Unfortunately, this is a breaking change for this library, which is unfortunate, as v0.29 not supporting older servers seems like a mistake to me. Should we solve this differently so that this can be released as a patch release? Or is it fine to just immediately release v0.30?


martin-g:
> Cargo supports having the same library multiple times with different versions if you need to handle multiple meilisearch versions in the same program

Can you explain more how this could work ?
How in my code I could say "use some_library version X" ?

Mubelotix:
This probably needs a few changes to compile but it will give you an idea

```toml
[dependencies]
meilisearch-sdk = { package = "meilisearch-sdk", version = "0.29" }
old-meilisearch-sdk = { package = "meilisearch-sdk", version = "0.28" }
```

```rust
use meilisearch_sdk::Client;
use old_meilisearch_sdk::Client as OldClient;

let client = Client::new();
let old_client = OldClient::new();
```

